PR #88, opened 2022-06-18
doc-link parser / links for non-kit sections

FractalHQ:
Documentation links for Svelte and Advanced Svelte sections.


Rich-Harris:
Going to close this as the main Svelte docs are undergoing an overhaul that would break all these links in any case — we'll have to take another run at it some time when we can properly think through how to structure the data in a maintainable fashion

PR #97, opened 2022-09-23
Add advisory about content being out of date

geoffrich:
I've seen some people in the Discord help channels refer to this as the "SvelteKit tutorial", and evidently some YouTube tutorials recommend it. We have the existing "Work in progress" message at the top but I thought this would be more clear that it's not up-to-date.


geoffrich:
Example advisory:

image

PR #105, opened 2022-10-17
tutorial

dummdidumm:


content/tutorial/02-sveltekit/02-routing/02-layouts/README.md
```diff
@@ -0,0 +1,32 @@
...
+title: Layouts
+---
+
+Sections of your app often share a common UI. To not repeat that code, we can use layouts.
+
+In this app we have two routes — `src/routes/+page.svelte`, and `src/routes/about/+page.svelte`. Both contain the same header which enables us to navigate around.
+
+We can deduplicate this code by moving the common UI into a `+layout.svelte` file. All pages in the same folder or below this file will share that UI. The layout itself needs a `<slot />` at minimum to define where the content is projected.
+
+Let's move the duplicate content from `src/routes/+page.svelte` and `src/routes/about/+page.svelte` into `src/routes/+layout.svelte`:
```
tanhauhau, line 11:
```suggestion
Let's move the common header from `src/routes/+page.svelte` and `src/routes/about/+page.svelte` into `src/routes/+layout.svelte`:
```

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Loading Data
+---
+
+So far, we have seen how to create pages and layouts with static UI data. Most of your apps need to load some kind of data however. You could do this with a regular `fetch` inside `onMount`. That doesn't work with server-side-rendering or prerendering though, and you'd have half-finished pages loading data only after they are rendered.
```
tanhauhau, line 5:
```suggestion
So far, we have seen how to create pages and layouts with static UI data. Most of your apps need to load some kind of data. You could do this with a regular `fetch` inside `onMount`. However, that doesn't work with server-side-rendering or prerendering, and you'd end up with half-finished pages that load data only after they are rendered.
```

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+SvelteKit provides a better way to do this through `load` functions.
+
+This `load` function can appear in several places. Let's start with the `load` function inside `+page.js`.
```
tanhauhau, line 9:
```suggestion
This `load` function can appear in several places. Let's start with adding the `load` function inside `+page.js`.
```

content/tutorial/02-sveltekit/02-routing/03-loading-data/app-a/src/routes/api/+server.js
```diff
@@ -0,0 +1,5 @@
+// We'll get to +server.js in a later chapter
+
+export function GET() {
+	return new Response('Hello from your API');
```
tanhauhau, line 4: im thinking maybe add a 5 second delay, so, we can show to the user that it is obvious that the page loads with a `<p>undefined</p>` then 5 seconds later turn into data.

as compared to takes 5 seconds to see the complete rendered page.

what do you think?

dummdidumm: I think that would make the SvelteKit-version look worse ("oh I have to wait 5 seconds until _anything_ is visible?") - I can see `undefined` flashing in the output, so I'm ok with that.

content/tutorial/02-sveltekit/02-routing/04-loading-layout-data/README.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Loading Layout Data
+---
+
+Layouts can provide common UI across pages. They can also provide common data across pages.
+
+Providing this data is similar to the `load` function in `+page.js`, only that we place it inside `+layout.js` - notice the symmetry?
+
+Like in `+page.js`, this `load` function runs on the server when the user visits your page initially, and on the client for all further navigations - it's isomorphic. We call it a "shared `load` function". To harmonize the differences between client and server, use the `fetch` function passed to `load` to load data. The return type of the `load` function needs to be an object at the top level, below that everything is allowed. If we return a promise, it's awaited. The result is passed to the `data` prop inside `+layout.svelte`, so we can access it through `export let data`.
```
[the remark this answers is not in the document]
dummdidumm: I did it on purpose in case anyone comes from Google and hasn't done the previous chapter - but yeah for someone doing it linearly it's duplicated. Not sure.

content/tutorial/02-sveltekit/02-routing/03-loading-data/app-a/src/routes/api/+server.js
tanhauhau: fair enough

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
dummdidumm: To me the current sentence feels more correct - it's not about the first step of what to do in this specific tutorial section but rather that we start with the load function inside this file.

content/tutorial/02-sveltekit/03-mutating-data/01-forms/app-b/src/routes/+page.svelte
```diff
@@ -0,0 +1,7 @@
+<p>Please log in</p>
+
+<form method="POST">
+	<input name="email" type="email" />
+	<input name="password" type="password" />
+	<button>Log in</button>
```
geoffrich, line 6: For accessibility, we should give these proper `<label>`s

content/tutorial/02-sveltekit/03-mutating-data/04-progressive-enhancement/README.md
```diff
@@ -0,0 +1,31 @@
...
+		<span>{form?.message}</span>
+	{/if}
+	<button>Log in</button>
+	<button formAction="?/register">Register</button>
+</form>
+```
+
+And that's all! Now, when JavaScript is enabled, `use:enhance` will emulate the browser-native behaviour, just without the full-page reloads. It will:
+
+- update the `form` property - but only if the action is on the same page you're submitting from. So for example if your form looks like <form action="/somewhere/else" ..>, form and $page will not be updated. This is because in the native form submission case you would be redirected to the page the action is on.
```
geoffrich, line 28: Possibly use code formatting here:

```suggestion
- update the `form` property - but only if the action is on the same page you're submitting from. So for example if your form looks like `<form action="/somewhere/else" ..>`, `form` and `$page` will not be updated. This is because in the native form submission case you would be redirected to the page the action is on.
```

content/tutorial/02-sveltekit/03-mutating-data/01-forms/app-b/src/routes/+page.svelte
geoffrich: (This applies to the rest of the examples in the form section as well)

geoffrich (commented):
Took a quick look, here are a couple suggestions. Looking good!

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
```diff
@@ -8,4 +8,8 @@ A SvelteKit app can be thought of as two distinct entities working in tandem —
 
 'Client' refers to the JavaScript that loads in the browser.
 
-TODO
+SvelteKit makes the two communicate with each other seamlessly. It takes care of all the annoying stuff so you don't have to.
+
+On first page hit, SvelteKit will render the HTML on the server and send it to the browser. This means content is visible as fast as possible to the user. SvelteKit will then take over that HTML in a process called hydration, so that subsequent navigations will happen in the browser, allowing for a better user experience.
```
Rich-Harris, line 13: we can probably afford to be more 'just the facts, ma'am' in the tutorial

```suggestion
SvelteKit makes the two communicate with each other seamlessly. On the initial page load, the server renders the HTML, meaning content is visible as quickly as possible. The client then takes over in a process called 'hydration', so that subsequent navigations happen without full page reloads. It will request additional code and data from the server as needed.
```

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
```diff
@@ -8,4 +8,8 @@ A SvelteKit app can be thought of as two distinct entities working in tandem —
...
+
+> You can adjust this behavior to your liking if you want - SvelteKit is very versatile!
```
Rich-Harris, line 15:
```suggestion
> You can [adjust this behavior](https://kit.svelte.dev/docs/page-options) as needed. SvelteKit is very versatile!
```

content/tutorial/02-sveltekit/01-concepts/04-server-side-rendering/README.md
```diff
@@ -2,4 +2,6 @@
 title: Server-side rendering
 ---
 
-TODO
+Server-side rendering (SSR) is the generation of the page contents on the server. It's what SvelteKit does by default when a user first visits your site.
```
Rich-Harris, line 5: the 'first' could be interpreted as 'the first time ever', i.e. subsequent visits are controlled by a service worker with an app shell, PWA-style

```suggestion
Server-side rendering (SSR) is the generation of HTML on the server. It's what SvelteKit does by default for the initial page load when a user visits your site.
```

content/tutorial/02-sveltekit/01-concepts/04-server-side-rendering/README.md
```diff
@@ -2,4 +2,6 @@
...
+
+SSR is generally preferred for SEO (search engine optimization). While some search engines can index content that is dynamically generated on the client-side it may take longer even in these cases. It also tends to improve perceived performance and makes your app accessible to users if JavaScript fails or is disabled (which happens [more often than you probably think](https://kryogenix.org/code/browser/everyonehasjs.html)).
```
Rich-Harris, line 7: I think we should put performance and resilience ahead of SEO, since a lot of people aren't concerned with SEO (but everyone should be concerned with the other two)

```suggestion
SSR tends to improve performance and makes your app accessible to users if JavaScript fails or is disabled (which happens [more often than you probably think](https://kryogenix.org/code/browser/everyonehasjs.html)). It is also preferred for search engine optimization (SEO) — while some search engines can index content that is rendered in the browser with JavaScript, it happens less frequently and reliably.
```

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
Rich-Harris: i guess there's an interesting conversation to be had here about whether tutorial sections should link to other tutorial sections, and how we want to approach structured two-way links between them

content/tutorial/02-sveltekit/01-concepts/05-prerendering/README.md
```diff
@@ -2,4 +2,10 @@
 title: Prerendering
 ---
 
-TODO
+Prerendering means computing the contents of a page at build time and saving the HTML for display. You can opt in to prerendering on a per-page-basis or even the whole app when using SvelteKit.
```
Rich-Harris, line 5:
```suggestion
Prerendering means computing the contents of a page at build time and saving the HTML. You can opt in to prerendering for a single page, the entire app, or anything in between.
```

content/tutorial/02-sveltekit/01-concepts/05-prerendering/README.md
```diff
@@ -2,4 +2,10 @@
...
+
+This approach has the same benefits as traditional server-rendered pages, but avoids recomputing the page for each visitor and so scales nearly for free as the number of visitors increases. The tradeoff is that the build process is more expensive and prerendered content can only be updated by building and deploying a new version of the application.
```
Rich-Harris, line 7:
```suggestion
This approach has the same benefits as traditional server-rendered pages, but avoids recomputing the page for each visitor and so scales nearly for free as the number of visitors increases. The tradeoff is that the build process takes longer, and prerendered content can only be updated by building and deploying a new version of the application.
```

content/tutorial/02-sveltekit/01-concepts/05-prerendering/README.md
```diff
@@ -2,4 +2,10 @@
...
+
+Not all pages can be prerendered. The basic rule is this: for content to be prerenderable, any two users hitting it directly must get the same content from the server, and the page must not contain actions. Note that you can still prerender content that is loaded based on the page's parameters as long as all users will be seeing the same prerendered content.
```
Rich-Harris, line 9: should `form actions` link to anything? People won't know what they are at this stage. (age old problem of cyclical dependencies in documentation)

```suggestion
Not all pages can be prerendered. The basic rule is this: for content to be prerenderable, any two users hitting it directly must get the same content from the server, and the page must not contain form actions. (Note that you can still prerender content that uses dynamic route parameters, as long as it meets those constraints.)
```

content/tutorial/02-sveltekit/01-concepts/05-prerendering/README.md
```diff
@@ -2,4 +2,10 @@
...
+
+Pre-rendered pages are not limited to static content. You can build personalized pages if user-specific data is fetched and rendered client-side. This is subject to the caveat that you will experience the downsides of not doing SSR for that content as discussed above.
```
Rich-Harris, line 11:
```suggestion
You can add user-specific data to a prerendered page by fetching it from the browser instead of using SvelteKit's built-in data loading, though this is generally not recommended since it will cause content to pop in as it loads.
```

Rich-Harris (commented):
man, this is an incredible amount of work — massive kudos. will do several mini-reviews rather than one epic one

content/tutorial/02-sveltekit/02-routing/02-layouts/app-a/src/routes/about/+page.svelte
```diff
@@ -0,0 +1,17 @@
...
+</nav>
+
+<h1>About</h1>
+
+<p>This is the about page.</p>
+
+<style>
+	nav {
+		display: flex;
+		justify-content: space-between;
```
Rich-Harris, line 13:
```suggestion
		gap: 1em;
```

content/tutorial/02-sveltekit/02-routing/02-layouts/app-b/src/routes/+layout.svelte
```diff
@@ -0,0 +1,15 @@
...
+	<a href="/">Home</a>
+	<a href="/about">About</a>
+</nav>
+
+<slot />
+
+<style>
+	nav {
+		display: flex;
+		justify-content: space-between;
```
Rich-Harris, line 11:
```suggestion
		gap: 1em;
```

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Loading Data
```
Rich-Harris, line 2:
```suggestion
title: Loading data
```

content/tutorial/02-sveltekit/02-routing/04-loading-layout-data/README.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Loading Layout Data
```
Rich-Harris, line 2:
```suggestion
title: Loading layout data
```

content/tutorial/02-sveltekit/02-routing/06-loading-layout-server-data/README.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Loading Layout Data on the Server
```
Rich-Harris, line 2: I don't think we need a separate section here, we can have

* loading page data
* loading layout data
* loading data (page _or_ layout) on the server

content/tutorial/02-sveltekit/02-routing/07-params/README.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Route Parameters
```
Rich-Harris, line 2: If we do the blog example, we should probably move this section above 'loading data' since it'll need parameters. I think we can make this section work without a `load` function — we can just show that if you have a `/blog/[slug]` route it will respond to any `/blog/*` pathname, and that leads naturally to talking about loading data in the next section

```suggestion
title: Route parameters
```

content/tutorial/02-sveltekit/02-routing/08-endpoints/README.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Endpoints
```
Rich-Harris, line 2: Do we want to use the 'endpoints' terminology? Every other framework calls them 'API routes'. I'm not 100% sure if that's better though. I generally refer to them as '`+server.js` routes' since that's unambiguous and cements the relationship between the two concepts, but that's probably a little unwieldy. Server routes?

content/tutorial/02-sveltekit/03-mutating-data/02-form-validation/app-b/src/routes/+page.server.js
```diff
@@ -0,0 +1,11 @@
+import { redirect } from '@sveltejs/kit';
```
Rich-Harris, line 1:
```suggestion
import { redirect, invalid } from '@sveltejs/kit';
```

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
```diff
@@ -4,13 +4,28 @@ title: Pages
...
 SvelteKit uses filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
 
-In this app we have two routes — `src/routes/+page.svelte`, which maps to `/`, and `src/routes/about/+page.svelte`, which maps to `/about`. Clicking the `about` link will take you from the home page to the about page.
+The routes are located within `src/routes`. Every directory within which contains a `+page.svelte` file creates a route in your app.
 
-> Unlike traditional multi-page apps, navigating to `/about` updates the contents of the current page, like a single-page app. This gives us the best of both worlds — fast server-rendered startup, then instant navigation.
+In this app we currently have one route — `src/routes/+page.svelte`, which maps to `/`.
 
-Let's add a link in `src/routes/about/+page.svelte` back to the homepage:
+Let's add a second route, `src/routes/about/+page.svelte`, which maps to `/about`:
```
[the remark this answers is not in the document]
dummdidumm: What do you mean? You should be able to left click and a context menu appears so you can add those files/folders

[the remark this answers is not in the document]
Rich-Harris: nope. left click where?

relatedly, i did wonder if a constrained way to allow the user to create new files would be if the auto-generated links like in the left-hand column like this one...

image

...created the file if it didn't already exist. maybe that's a confusing UI, but i figured I'd throw it out there

Rich-Harris:
we'll have to figure out what to do in this situation:

image

either way need to hide the `` ` `` character or we need to turn it into `<code>`. i prefer the first option, `<code>` will be a CSS nightmare

content/tutorial/02-sveltekit/02-routing/05-loading-server-data/README.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Loading Data on the Server
```
Rich-Harris, line 2:
```suggestion
title: Loading data on the server
```

content/tutorial/02-sveltekit/03-mutating-data/02-form-validation/README.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Form Validation
```
Rich-Harris, line 2:
```suggestion
title: Form validation
```

content/tutorial/02-sveltekit/03-mutating-data/03-named-form-actions/README.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Named Form Actions
```
Rich-Harris, line 2:
```suggestion
title: Named form actions
```

content/tutorial/02-sveltekit/03-mutating-data/04-progressive-enhancement/README.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Progressive Enhancement
```
Rich-Harris, line 2:
```suggestion
title: Progressive enhancement
```

content/tutorial/02-sveltekit/03-mutating-data/06-mutation-through-endpoints/README.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Using Endpoints for Data Mutation
```
Rich-Harris, line 2:
```suggestion
title: Using endpoints for data mutation
```

content/tutorial/02-sveltekit/04-errors-and-redirects/01-error-pages/README.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Error Pages
```
Rich-Harris, line 2:
```suggestion
title: Error pages
```

content/tutorial/02-sveltekit/06-rendering-options/01-rendering-options-basics/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Rendering Options Basics
```
Rich-Harris, line 2:
```suggestion
title: Basics
```

content/tutorial/02-sveltekit/06-rendering-options/02-ssr/README.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Server Side Rendering
```
Rich-Harris, line 2:
```suggestion
title: Server-side rendering
```

content/tutorial/02-sveltekit/06-rendering-options/03-csr/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Client Side Rendering
```
Rich-Harris, line 2:
```suggestion
title: Client-side rendering
```

content/tutorial/04-advanced-sveltekit/01-advanced-routing/01-optional-params/README.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Optional Parameters
```
Rich-Harris, line 2:
```suggestion
title: Optional parameters
```

content/tutorial/04-advanced-sveltekit/01-advanced-routing/02-rest-params/README.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Rest Parameters
```
Rich-Harris, line 2:
```suggestion
title: Rest parameters
```

content/tutorial/04-advanced-sveltekit/01-advanced-routing/03-param-matchers/README.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Param Matchers
```
Rich-Harris, line 2:
```suggestion
title: Param matchers
```

content/tutorial/04-advanced-sveltekit/01-advanced-routing/04-layout-groups/README.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Layout Groups
```
Rich-Harris, line 2:
```suggestion
title: Layout groups
```

Rich-Harris (commented):
figured i'd do all these in one go

content/tutorial/02-sveltekit/01-concepts/04-server-side-rendering/README.md
```diff
@@ -2,4 +2,6 @@
 title: Server-side rendering
 ---
 
-TODO
+Server-side rendering (SSR) is the generation of HTML on the server. It's what SvelteKit does by default for the initial page load when a user visits your site.
+
+SSR tends to improve performance and makes your app accessible to users if JavaScript fails or is disabled (which happens [more often than you probably think](https://kryogenix.org/code/browser/everyonehasjs.html)). It is also preferred for search engine optimization (SEO) — while some search engines can index content that is rendered in the browser with JavaScript, it happens less frequently and reliably.
```
Rich-Harris, line 7: I kinda wonder if we should fold this and the subsequent 'Prerendering' section into the previous 'Server and client' section, or at least combine them into a 'Rendering modes' section (that includes `csr`?).

Then again, all this is somewhat duplicative of content that will appear later, so I almost wonder if we should ruthlessly cut down the 'concepts' section.

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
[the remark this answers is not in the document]
dummdidumm: Oh woops I meant right click, in the file tree on one of the folders or files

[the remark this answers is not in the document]
Rich-Harris: Ah okay it works on folders — I could have sworn I tried that. I think the context menu is a little undiscoverable — what if icons appeared in the right margin on hover?

content/tutorial/02-sveltekit/02-routing/04-loading-layout-data/README.md
```diff
@@ -0,0 +1,37 @@
...
+  +layout.svelte
+  +page.svelte
+```
+
+Then add the `load` function, calling the API inside, and passing the result to `+layout.svelte` for display:
+
+```js
++++export async function load({ fetch }) {
+	const greeting = await fetch('/api').then((r) => r.text());
+	return { greeting };
```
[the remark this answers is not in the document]
dummdidumm: `await parent()` is part of the advanced SvelteKit section, I'm not sure I'd want to introduce it here

content/tutorial/02-sveltekit/02-routing/06-loading-layout-server-data/README.md
dummdidumm: Something like "we can also move these calls to the server an then safe us the indirection of going through our own `+server.js`, let's rename `+page.js` to `+page.server.js` and `+layout.js` to `+layout.server.js` and adjust the load functions to call the fake database directly?

content/tutorial/02-sveltekit/02-routing/04-loading-layout-data/README.md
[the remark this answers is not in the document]
Rich-Harris: Come to think of it, I wonder if it's better to start with server-only `load` functions and _then_ cover shared `load` functions in a separate example — perhaps something loading data from an external API (preferably one we trust not to fall over)

content/tutorial/02-sveltekit/02-routing/08-endpoints/README.md
dummdidumm: If every other framework does then, I'm fine with API routes. I just know this as "endpoints" from my previous job.

content/tutorial/02-sveltekit/02-routing/06-loading-layout-server-data/README.md
Rich-Harris: We shouldn't be talking about `+server.js` here at all, it's unnecessarily confusing. Re order, see https://github.com/sveltejs/learn.svelte.dev/pull/105/files#r1038471015

content/tutorial/02-sveltekit/06-rendering-options/01-rendering-options-basics/README.md
dummdidumm: I think it's better to go with just "rendering options" instead.

content/tutorial/02-sveltekit/01-concepts/04-server-side-rendering/README.md
dummdidumm: I was wondering the same above already - I'll take this as a "yes, remove it"

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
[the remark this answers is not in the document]
dummdidumm: It's restricted per tutorial - you can only add/delete the files/folders you need for the tutorial to complete, that's why in that case you don't get it for files, because there's no need to rename/delete a file in that chapter. Making it more discoverable: I was expecting it to feel natural because that's what you do in the regular explorer view on your desktop, too. But it's probably not in the tutorial if you start with Svelte tutorial chapters where there's no possibility to add/remove anything.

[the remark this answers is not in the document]
benmccann: I had no trouble with the file tree. My one suggestion for it was going to be if we could add vertical lines showing which folder a file/folder was in (VS Code does this. you also do it right below when you type it out with `├` and `|` characters)

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
```diff
@@ -4,13 +4,28 @@ title: Pages
...
++│ └ +page.svelte
+└ +page.svelte
+```
+
+Clicking the `about` link on the home page will now take you to the about page (which is empty right now).
+
+Let's add some content (including a link back to the homepage) to `src/routes/about/+page.svelte` :
 
 ```svelte
++++<h1>About</h1>
```
benmccann, line 25: I got a popup when I tried to copy/paste this, but checking the box didn't seem to enable it. I would find it super annoying if I had to type all this and just gave up and hit solve

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
```diff
@@ -4,13 +4,28 @@ title: Pages
...
+
 <p>This is the about page.</p>
-<p>Go to the +++<a href="/">+++home+++</a>+++ page</p>
+<p>Go to the <a href="/">home</a> page</p>+++
 ```
+
+> Unlike traditional multi-page apps, navigating to `/about` and back updates the contents of the current page, like a single-page app. This gives us the best of both worlds — fast server-rendered startup, then instant navigation.
```
benmccann, line 31: perhaps mention this is the default behavior, but configurable

content/tutorial/02-sveltekit/02-routing/02-layouts/README.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Layouts
+---
+
+Different routes of your app will often share common UI. Instead of repeating it in each `+page.svelte` component, we can use a `+layout.svelte` component that applies to all routes in the same directory.
```
benmccann, line 5:
```suggestion
Different routes of your app will often share common UI. Instead of repeating the common portions in each `+page.svelte` component, we can use a `+layout.svelte` component that applies to all routes in the same directory.
```

dummdidumm:
> we'll have to figure out what to do in this situation:
> 
> image
> 
> either way need to hide the `` ` `` character or we need to turn it into `<code>`. i prefer the first option, `<code>` will be a CSS nightmare

I just removed the backticks, they aren't necessary and not styled either in the title

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+So far, we have seen how to create pages and layouts with static UI data. Most of your apps need to load some kind of data. You could do this with a regular `fetch` inside `onMount`. However, that doesn't work with server-side-rendering or prerendering, and you'd end up with half-finished pages that load data only after they are rendered.
+
+SvelteKit provides a better way to do this through `load` functions.
```
benmccann, line 7: I think it might be nice to link to https://kit.svelte.dev/docs/load here

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This `load` function runs on the server when the user visits your page initially, and on the client for all further navigations - it's isomorphic. We call it a "shared `load` function". To make sure it works the same way in the client and server, use the `fetch` function passed into `load` to fetch data.
+
+The return type of the `load` function needs to be an object at the top level, below that everything is allowed. If we return a promise, it's awaited. The result is passed to the `data` prop inside `+page.svelte`, so we can access it through `export let data`.
+
+Let's move the call inside `src/routes/+page.svelte` into `+page.js`. First, create the file:
+
+```diff
+src/routes/
++  +page.js
```
benmccann, line 19: we're missing the `├` characters here that were on earlier pages. I think it was nice having them

content/tutorial/02-sveltekit/02-routing/05-loading-server-data/README.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+We have seen how to load data for pages and layouts through `load` functions in `+page.js` and `+layout.js`. These functions run on both the server and the client. This has some advantages but also some drawbacks, for example you can't directly query your database, you have to use `fetch`.
+
+Luckily, SvelteKit provides a way to run a `load` _only_ on the server so you can access all your server-related data directly, by placing it inside `+page.server.js`. We call it a "server-only `load` function". The return type of the `load` function needs to be an object at the top level, and it needs to be serializable with `devalue`, which means JSON-objects and some JavaScript objects like `Date`s are allowed, but not custom classes for example. If we return a promise, it's awaited. The result is passed to the `data` prop inside `+page.svelte`, so we can access it through `export let data`.
```
benmccann, line 7:
```suggestion
Luckily, SvelteKit provides a way to run a `load` _only_ on the server so you can access all your server-related data directly, by placing it inside `+page.server.js`. We call it a "server-only `load` function". The return type of the `load` function needs to be an object at the top level, and it needs to be serializable with [`devalue`](https://www.npmjs.com/package/devalue), which means JSON-objects and some JavaScript objects like `Date`s are allowed, but not custom classes for example. If we return a promise, it's awaited. The result is passed to the `data` prop inside `+page.svelte`, so we can access it through `export let data`.
```

content/tutorial/02-sveltekit/02-routing/06-loading-layout-server-data/README.md
dummdidumm: That sounds good. I was thinking about the blog example initially, too, but felt it may be too complex and hide the actual teachings. But it sounds like you have a good idea for a simple version. Feel free to shuffle the routing chapter around (although @benmccann wanted to review, too, now, maybe he has a different opinion), getting late here.

benmccann (commented):
wow, this is looking great! such great progress!

I'd really like if the menu showed all the pages. E.g. it just says "Errors And Redirects", but there are many different pages within that section and I want to see how far I've progressed, be able to jump back and forth between an old page and the current page, etc.

When you're on /tutorial/prerender, which is the very last page of Part 2, the button at the bottom of the page just says "Next: Tweens", but should probably say something more like "Next: Part 3: Advanced Svelte"

I would make the left pane with tutorial text and file explorer both a little larger by default.

We should have some kind of "congratulations! you're done!" page at the end maybe with svelte confetti.

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
dummdidumm: mhm that's strange, I definetely checked this box multiple times already 😄

content/tutorial/02-sveltekit/02-routing/04-loading-layout-data/README.md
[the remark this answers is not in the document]
Rich-Harris: > `await parent()` (which we would need to "has access to layout data") is part of the advanced SvelteKit section, I'm not sure I'd want to introduce it here

You can access `data.posts` inside the component. We don't need it inside the page's `load` function

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
```diff
@@ -8,4 +8,6 @@ A SvelteKit app can be thought of as two distinct entities working in tandem —
 
 'Client' refers to the JavaScript that loads in the browser.
 
-TODO
+SvelteKit makes the two communicate with each other seamlessly. On the initial page load, the server renders the HTML, meaning content is visible as quickly as possible. The client then takes over in a process called 'hydration', so that subsequent navigations happen without full page reloads. It will request additional code and data from the server as needed.
```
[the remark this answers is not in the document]
Rich-Harris: 'server the server-side'... 😬 

The confusingness doesn't just come from serverless/edge functions, but also prerendering (and one day, service workers)

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
Rich-Harris: You have to copy again after enabling it. Maybe we can write the content to the clipboard as soon as it's enabled

content/tutorial/02-sveltekit/02-routing/03-loading-data/README.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Loading data
+---
+
+So far, we have seen how to create pages and layouts with static UI data. Most of your apps need to load some kind of data. You could do this with a regular `fetch` inside `onMount`. However, that doesn't work with server-side-rendering or prerendering, and you'd end up with half-finished pages that load data only after they are rendered.
+
+SvelteKit provides a better way to do this through `load` functions.
```
Rich-Harris, line 7:
```suggestion
SvelteKit provides a better way to do this through [`load`](https://kit.svelte.dev/docs/load) functions.
```

Rich-Harris:
> I'd really like if the menu showed all the pages. E.g. it just says "Errors And Redirects", but there are many different pages within that section

@benmccann I'm not sure what this means? This is what I see 

image

benmccann:
Menu looks fine to me now. The sections weren't appearing expandable to me for some reason before if I recall

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
benmccann: Ah, yeah, I might not have realized that. Writing to the clipboard sounds nice

Rich-Harris:
had you searched? it filters by search term

image

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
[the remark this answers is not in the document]
benmccann: > 'serve the server-side'...

haha. yeah, that was bad. updated to remove that

> The confusingness doesn't just come from serverless/edge functions, but also prerendering

It sounds like the confusing part isn't really the _word_ "server"? Maybe we should just swap the word "confusing" with "apropos"?

> and one day, service workers

In that case we'd have to also reword the prior sentence as there's no longer "two distinction entities", but now three: the client, server, and service worker.

benmccann:
don't think so, but it was last week. in any case, it looks good to me now, so I wouldn't worry about it

content/tutorial/02-sveltekit/02-routing/01-pages/README.md
Rich-Harris: Unfortunately this isn't possible — we can't programmatically read and write to and from the clipboard unless the user has given permission, and we can't request permission and _also_ call `event.preventDefault()` because permissions are async.

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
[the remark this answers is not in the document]
Rich-Harris: > Maybe we should just swap the word "confusing" with "apropos"?

These are essentially opposites?

Again, among other things we're talking about prerendering here. In a literal sense there is no server at all, even a local server — it's just a function. But conceptually that function is a server.

In the service worker case, there are still two entities; the service worker _is_ the 'server'.

If there's a better word than 'server' to describe the abstract concept of 'thing that takes a `Request` as input and returns a `Response` then I'm all ears, but I don't think such a thing exists. Absent one, our best approach is to use the common terminology but be upfront about its shortcomings.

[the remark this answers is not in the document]
benmccann: I should have been clearer. By swap, I didn't mean a 1:1 swap, but rewriting to make it work with that word. Like:

>'Server' is, perhaps, not the most apropos word since your app will often be running in a _serverless_ environment (cloud/edge functions) or might even be deployed as a set of completely static files. But it's the best we've got. The server's basic job is to turn a request into a response.

dummdidumm:
I see you renamed "rendering options" to "page options" - which reminds me that I wanted to advocate for changing the section to be named "rendering options" in the kit docs, too. "page options" to me sounds like it's tied to a page, but you can also set those on layouts. Moreover, they all relate to rendering, so I think that "rendering options" is the more fitting term.

content/tutorial/02-sveltekit/03-loading-data/01-page-data/README.md
```diff
@@ -0,0 +1,94 @@
...
++++<script>
+	export let data;
+</script>+++
+
+---<h1>Blog post</h1>---
++++<h1>{data.post.title}</h1>
+<div>{@html data.post.content}</div>+++
+```
+
+There's one last detail we need to take care of — the user might visit an invalid pathname like `/blog/nope`, in which case we'd like to respond with a 404 page:
```
dummdidumm, line 76: Do we really need this part of the tutorial? If we just gloss over this for now, almost noone would notice, and it's one less thing to worry about for now. Furthermore, this is pretty much all there is to explain about expected errors, so the later section where this is explained would feel rather duplicative.

(would also give room to add that "notice how we can colocate our database stuff" info box I wrote previously, which I thought is a nice argument in favor of our +convention)

content/tutorial/02-sveltekit/03-loading-data/01-page-data/app-b/src/routes/blog/[slug]/+page.server.js
```diff
@@ -0,0 +1,12 @@
+import { error } from '@sveltejs/kit';
+import { posts } from '../data.js';
+
+export function load({ params }) {
+	const post = posts.find((post) => post.slug === params.slug);
+
+	if (!post) throw error(404);
```
Comment (dummdidumm) [inline content/tutorial/02-sveltekit/03-loading-data/01-page-data/app-b/src/routes/blog/[slug]/+page.server.js:7]: similar to above comment

content/tutorial/02-sveltekit/03-loading-data/02-layout-data/app-a/src/routes/blog/[slug]/+page.server.js
```diff
@@ -0,0 +1,12 @@
+import { error } from '@sveltejs/kit';
+import { posts } from '../data.js';
+
+export function load({ params }) {
+	const post = posts.find((post) => post.slug === params.slug);
+
+	if (!post) throw error(404);
```
Comment (dummdidumm) [inline content/tutorial/02-sveltekit/03-loading-data/02-layout-data/app-a/src/routes/blog/[slug]/+page.server.js:7]: similar to above comment

content/tutorial/02-sveltekit/03-loading-data/02-layout-data/app-b/src/routes/blog/[slug]/+page.server.js
```diff
@@ -0,0 +1,12 @@
+import { error } from '@sveltejs/kit';
+import { posts } from '../data.js';
+
+export function load({ params }) {
+	const post = posts.find((post) => post.slug === params.slug);
+
+	if (!post) throw error(404);
```
Comment (dummdidumm) [inline content/tutorial/02-sveltekit/03-loading-data/02-layout-data/app-b/src/routes/blog/[slug]/+page.server.js:7]: similar to above comment

content/tutorial/02-sveltekit/03-loading-data/03-universal-load-functions/app-a/src/routes/blog/+page.server.js
```diff
@@ -0,0 +1,15 @@
+import { error } from '@sveltejs/kit';
+
+export async function load({ fetch }) {
+	const response = await fetch('/api/blog');
+
+	if (!response.ok) {
+		throw error(response.status);
```
dummdidumm, line 7: similar to above comment

content/tutorial/02-sveltekit/03-loading-data/03-universal-load-functions/app-a/src/routes/blog/[slug]/+page.server.js
```diff
@@ -0,0 +1,11 @@
+import { error } from '@sveltejs/kit';
+
+export async function load({ fetch, params }) {
+	const response = await fetch(`/api/blog/${params.slug}`);
+
+	if (!response.ok) {
+		throw error(response.status);
```
Comment (dummdidumm) [inline content/tutorial/02-sveltekit/03-loading-data/03-universal-load-functions/app-a/src/routes/blog/[slug]/+page.server.js:7]: similar to above comment

dummdidumm (commented):
(comments about maybe leaving out `throw error` out of this section)

content/tutorial/02-sveltekit/06-errors-and-redirects/01-error-pages/README.md
```diff
@@ -0,0 +1,38 @@
...
+		: true;
+</script>
+
+{#if $page.status === 404}
+	<h1>Not found</h1>
+{:else if !online}
+	<h1>You're offline</h1>
+{:else}
+	<h1>Oops!</p>
+	<p>{$page.error.message}</p>
```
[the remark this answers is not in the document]
dummdidumm: (did that in a commit and adjusted the following section)

content/tutorial/02-sveltekit/03-loading-data/01-page-data/README.md
Rich-Harris: error handling is like a11y — it's good to show best practices early, even if it means jumping ahead. people will 100% copy this code and ship buggy apps if we don't nudge people in the right direction straight away

content/tutorial/02-sveltekit/01-concepts/03-server-and-client/README.md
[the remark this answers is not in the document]
Rich-Harris: I still think

> 'Server' is, perhaps, a confusing word

is better — 'not the most apropos word' implies that there's a _more_ apropos word, in which case why aren't we using it? The fact is that there _isn't_ a more apropos word. We should just acknowledge the fact and move on, I think

benmccann:
Do we want to remove "WARNING: this content is a work in progress and may be out of date" at the beginning?

I figured out the issue I was having with the menu last week as I just hit it again. I expected to be able to click the `>` button and thought the menu was broken or didn't expand when it didn't do anything. If you hadn't told me earlier that the menu expanded, I'm not sure I would have figured it out this time around

Also, the menu isn't expanded to show the current page. E.g. I'm currently on "Part 1 / Bindings / Group inputs" and the section that's expanded is the one above it: "Events"

The "new file" vs "new folder" icons are nearly indistinguishable to me. I had to move way closer to the monitor to see them. Maybe we could make them slightly bigger?

content/tutorial/04-advanced-sveltekit/06-environment-variables/01-private/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `+page.server.js`
+- `+layout.server.js`
+- `+server.js`
+- any modules ending with `.server.js`
+- any modules inside `src/lib/server`
+
+In turn, these modules can only be imported by _other_ server modules.
+
+## Static vs dynamic
```
dummdidumm, line 79: I would move this whole section into a follow-up exercise (I liked the previous distinction into 4 exercises instead of now 2 better), the tutorial gets too long at this point, and the "show me" skips the static imports version.

scripts/create-common-bundle/index.js
```diff
@@ -20,9 +20,8 @@ const ignored_extensions = ['.d.ts', '.map'];
 const ignored_directories = ['.svelte-kit', 'node_modules/.bin', 'node_modules/rollup/dist/shared'];
 
 const ignored_files = new Set([
-	'node_modules/rollup/dist/es/rollup.browser.js',
-	'node_modules/rollup/dist/rollup.browser.js',
-	'node_modules/svelte/compiler.js'
+	'node_modules/svelte/compiler.js',
+	'node_modules/@esbuild/darwin-arm64/bin/esbuild'
```
dummdidumm, line 24: why did you add this? It's failing the build.

content/tutorial/04-advanced-sveltekit/03-advanced-routing/01-optional-params/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+> TODO the filesystem constraints are broken here
+
+In the first chapter on [routing](/tutorial/pages), we learned how to create routes with [dynamic parameters](/tutorial/params).
+
+Sometimes it's helpful to make a parameter optional. A classic example is when you use the pathname to determine the locale — `/fr/...`, `/de/...` and so on — but you also want to have a default locale.
+
+To do that, we use double brackets. Rename `src/routes/[lang]` to `src/routes/[[lang]]`.
+
+The app now fails to build, because `src/routes/+page.svelte` and `src/routes/[[lang]]/+page.svelte` would both match `/`. Delete `src/routes/+page.svelte`.
```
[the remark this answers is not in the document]
Rich-Harris: We should probably fix it in Kit, if we can figure out a sensible way.

The only alternative is to not have a `src/routes/+page.svelte` in the first place, which felt like a worse option. Besides, people will run into this issue when they start using optional params, so explaining why it's happening (and demoing the fact that we prevent you from having conflicting routes) doesn't seem like the worst thing

content/tutorial/04-advanced-sveltekit/03-advanced-routing/03-param-matchers/README.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Param matchers
+---
+
+To prevent the router from matching on invalid input, you can specify a _matcher_. For example, you might want a route like `/colors/[value]` to match hex values like `/colors/ff3e00` but not named colors like `/colors/octarine` or any other arbitrary input.
```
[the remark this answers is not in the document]
Rich-Harris: a) it's more fun — it introduces some visual variety, which these examples need, and b) this isn't a complete example:

image

We're not actually _using_ the `page` param anywhere — there's no data, and so nothing to paginate — so to me it feels terribly confusing. Ideally these exercises should feel like complete self-contained entities, without 'draw the rest of the owl' implications

content/tutorial/04-advanced-sveltekit/06-environment-variables/01-private/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+Open `src/routes/+page.server.js`. Import `PASSPHRASE` from `$env/static/private` and use it inside the [form action](/tutorial/the-form-element):
+
+```js
+/// file: src/routes/+page.server.js
+import { redirect, fail } from '@sveltejs/kit';
++++import { PASSPHRASE } from '$env/static/private';+++
+
+export function load({ cookies }) {
+	if (cookies.get('allowed')) {
```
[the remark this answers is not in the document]
Rich-Harris: I think we need an example where the environment variable is actually used for something though, otherwise I'm left with the same 'what am I actually doing here?' confusion as the `[page=integer]` example. I couldn't think of any examples (like database connection strings or external API keys) that were realistic to implement in the tutorial. 

It's a good thing if later exercises reinforce things learned in earlier ones — layering the material is the best way for people to develop familiarity (or remind them to revisit past exercises if they find themselves confused by what they're looking at). It reminds people that the `cookies` API is a thing, and it's a reminder of how to use form actions. Take the point about persistence — I added a `maxAge` of `60` for that reason, but we could either a) reduce that further or b) add a `logout` action on the `/welcome` page, which makes the example more complete in any case

scripts/create-common-bundle/index.js
Rich-Harris: huh, I didn't think it was necessary, I thought we only needed the wasm binary. worked locally. the bundle is kinda chunky, was trying to slim it down

dummdidumm: My guess is that this is what esbuild chooses for you locally to include but on Vercel it chooses something else. Maybe we need some kind of wildcard there.

content/tutorial/04-advanced-sveltekit/06-environment-variables/01-private/README.md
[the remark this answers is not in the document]
dummdidumm: I don't mean to remove that from the other load function where it errors if you're not "authenticated", only from the "enter the passphrase" page

Rich-Harris:
> Do we want to remove "WARNING: this content is a work in progress and may be out of date" at the beginning?

Done, though we will need to do a proper run-through to make sure that everything _is_ up to date 😆 

> I figured out the issue I was having with the menu last week as I just hit it again. I expected to be able to click the `>` button and thought the menu was broken or didn't expand when it didn't do anything

Not sure I follow?



> Also, the menu isn't expanded to show the current page. E.g. I'm currently on "Part 1 / Bindings / Group inputs" and the section that's expanded is the one above it: "Events"

Ah yeah, the state persists — whatever you had open before remains open. We should reset it whenever you open the menu

> The "new file" vs "new folder" icons are nearly indistinguishable to me. I had to move way closer to the monitor to see them. Maybe we could make them slightly bigger?

Can probably make them a smidge bigger but I wonder if we need to tweak the icons themselves. These are the VSCode icons...

image

...and these are ours:

image

I think if the folder icon was slightly shorter, and the page icon didn't have a fold (or at least had a smaller fold) then the icons would be a more recognisably different shape

scripts/create-common-bundle/index.js
Rich-Harris: ah okay, it installs a different binary on vercel (obviously, duh) — will need to change this to use a regex instead

benmccann:
>> I figured out the issue I was having with the menu last week as I just hit it again. I expected to be able to click the > button and thought the menu was broken or didn't expand when it didn't do anything
> Not sure I follow?

These things don't do anything when you click them. I expected them to expand the corresponding menu item and when it didn't I assumed the menu was broken

Screenshot from 2022-12-11 13-44-14

sw-yx:
i've been running thru today (better late than never!) and found that theres a small bug: when you toggle between "solve" and "reset" on some pages, the "reset" button sometimes fails to go back to "solve" https://www.loom.com/share/d5d18025f9fb4ddb98b4eec65b832db7

sw-yx:
another bug - this `src/error.html` customization doesnt work as advertised: https://learn-svelte-mt5f8s1ev-svelte.vercel.app/tutorial/fallback-errors

<details>
<summary> screenshot
</summary>

image

</details>

Rich-Harris:
yeah there's definitely lots of WIP stuff still — some broken/undesigned UI, some content that isn't done yet, missing default styles etc

Rich-Harris:
(which isn't to discourage bug reports!)

content/tutorial/04-advanced-sveltekit/03-advanced-routing/01-optional-params/README.md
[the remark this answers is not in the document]
dummdidumm: Fixed in Kit, should show an error overlay now

content/tutorial/04-advanced-sveltekit/03-advanced-routing/01-optional-params/README.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Optional parameters
+---
+
+> TODO the filesystem constraints are broken here
```
dummdidumm, line 5: what's broken here? I can rename and delete the files as expected

dummdidumm:
- Show me/reset state toggle fixed
- the `error.html` not triggering anything is a bug inside SvelteKit itself, we need to watch these files, I'll send a fix

content/tutorial/02-sveltekit/03-loading-data/03-universal-load-functions/README.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Universal load functions
+---
+
+> Coming soon
```
dummdidumm, line 5: This feels rather drastic, and a "coming soon" so early is no good first impression in my opinion. `+page/layout.js` is a pretty big part of SvelteKit, so not teaching it ... I don't know, that feels wrong. I think it would be better to keep it, or adjust the text and say "let's assume that this is an external API".

content/tutorial/04-advanced-sveltekit/04-advanced-loading/05-invalidate-all/README.md
```diff
@@ -0,0 +1,42 @@
...
+/// file: src/routes/[...timezone]/+page.svelte
+<script>
+	import { onMount } from 'svelte';
+	import { +++invalidateAll+++ } from '$app/navigation';
+
+	export let data;
+
+	onMount(() => {
+		const interval = setInterval(() => {
+			+++invalidateAll();+++
```
dummdidumm, line 19: is it on purpose that you only did the `+++invalidateAll()+++` and not the `---invalidate(..)---` (same question applies to the other invalidate exercises)?

Rich-Harris: it's just unnecessary — it's clear from context what you're supposed to remove. Having the red and the green butting against each other makes for visual chaos, and makes it harder to understand what you're meant to do IMHO

content/tutorial/02-sveltekit/03-loading-data/03-universal-load-functions/README.md
Rich-Harris: yeah, i intend to follow up with stackblitz ASAP about why `fetch` is failing in this context. but having magic hidden routes is just a huge no-no in my opinion

PR #115, opened 2022-12-05
Loading

Rich-Harris:
This updates the loading section per https://github.com/sveltejs/learn.svelte.dev/pull/105/files#r1038350603.

Couple of things to note:

* it uses the `/// file: some-file-name.js` convention from the Kit docs, but there's no CSS in place for that just yet, so it looks a bit funky
* the `event.fetch(...)` call in the Hacker News example doesn't work, because the response is truncated. I can only assume this is a problem with WebContainers somehow, but I haven't investigated yet
* We need a way to rename/delete files. Also, up till now `app-b` was assumed to be additive, but in reality there are times when `app-a` contains files that shouldn't exist when the exercise is solved

I think it's _probably_ worth merging the content into `tutorial-routing` and worrying about the functionality later


dummdidumm:
Deleting files is already possible. It's not that pretty, but if you create a folder with the file `__delete` in it, it will mark the path above it as " will be removed". So if you want to delete a file you would do `src/routes/+page.svelte/__delete` . You can see it 'in action' in the advanced SvelteKit tutorial

Rich-Harris:
What do you think about requiring `app-b` to be complete, and expecting any files that are in `app-a` but not `app-b` to have been deleted? Do we think that would create more work or less work in the long run?

dummdidumm:
I see that you added notes on `$types` - frankly, I would completely remove those references, and worry about that later. Right now, they don't work at all in the context of the tutorial, and it will inevitably lead to "but where's the TypeScript version" questions.

content/tutorial/02-sveltekit/02-routing/04-loading-page-data/README.md
```diff
@@ -0,0 +1,101 @@
...
+	/** @type {import('./$types).PageData} */
+	export let data;
+</script>+++
+
+---<h1>Blog post</h1>---
++++<h1>{data.post.title}</h1>
+<div>{@html data.post.content}</div>+++
+```
+
+There's one last detail we need to take care of — the user might visit an invalid pathname like `/blog/nope`, in which case we'd like to respond with a 404 page:
```
dummdidumm, line 82: Do we really need this part of the tutorial? If we just gloss over this for now, almost noone would notice, and it's one less thing to worry about for now. Furthermore, this is pretty much all there is to explain about expected errors, so the later section where this is explained would feel rather duplicative.

content/tutorial/02-sveltekit/02-routing/07-api-routes/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+	return new Response(number, {
+		'Content-Type': 'application/json'
+	});
+}
+```
+
+Clicking the button now works.
+
+Request handlers must return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) object. Since it's common to return JSON from an API route, SvelteKit provides a convenience function for generating these responses, complete with the appropriate `Content-Type` header:
```
dummdidumm, line 23:
```suggestion
Request handlers must return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) object. Since it's common to return JSON from an API route, SvelteKit provides a convenience function for generating these responses:
```

dummdidumm (commented):
Looks mostly good to me - great work! This feels a lot more fleshed out now. Left a few small comments, and one on the `$types` situation, which I think we should solve by leaving this out of the tutorial for now.

PR #137, opened 2022-12-14
show progress indicator

Rich-Harris:
adds a progress bar


foucist:
I am curious if you would ever consider making Safari your primary development browser, develop for Safari first, Firefox second, and Chromium-based last?  I know it's a bit of a crazy idea but I feel like the underdogs need more love.

PR #140, opened 2022-12-14
Dynamic Attributes example: make example HTML consistent

seangates:
Updated the example to add consistency between the README and the editor experience.


dummdidumm (approved):
Thank you!

PR #143, opened 2022-12-15
Update README.md

andyrosa:
Consistency with example above


dummdidumm (approved):
Makes sense, thank you!

PR #145, opened 2022-12-15
Fix app-a

apatheticcoder:
Learning module suppose to show shorthand vs longhand but has shorthand in both app-a and app-b.


dummdidumm (approved):
Thank you!
(this probably got in through prettier formatting 😄 )

PR #153, opened 2022-12-15
Fix Typo

jinyus:
should be bar and baz


Rich-Harris:
thanks! closing in favour of #144

PR #155, opened 2022-12-15
fix: <a> tag in footer to open in a new tab

mateusabelli:
If the content of the sidebar is long enough the users might misclick and choose the *Edit this page* link instead of the *Next: Lesson* link. By doing so, the link will open within the user lesson tab and even if they go back to their previous tab the code progress on that lesson will be lost.

This patch makes the <a> tag on the footer open in a new tab, so if the user misclicks on the link they can close it and continue normally.


dummdidumm (approved):
Makes sense, thank you!

PR #159, opened 2022-12-16
[fix] reset focus after iframe reload

### LINKED ISSUE #150: Text input focus sometimes lost on rerender/error
Just typing I find the text input focus is lost randomly which makes typing difficult. The following gif is from https://learn.svelte.dev/tutorial/your-first-component

focus

This seems like a problem because I'm writing the script tag and end tag before the contents of it, is that something I shouldn't do?

dummdidumm:
Fixes #150


Rich-Harris:
Unfortunately this didn't work for me — you can see in this video I lose focus after `let na`:

https://user-images.githubusercontent.com/1162160/208138476-686729a9-b9d4-4da0-8b0d-ac04620bc0de.mov

Sometimes the focus gets captured by the iframe outside `set_iframe_src`, in ways that we have no control over. I have a fix that seems to work, will open a new PR

dummdidumm:
Strange that you get it immediately and I couldn't get it anymore after my testing - but your PR is the more robust solution, so will close in favor of that.

PR #160, opened 2022-12-16
Removed unused variable in tutorial/adding-parameters-to-actions

emircanerkul:
Removed unused variable in tutorial/adding-parameters-to-actions 's first view. It should be placed after solving the problem


dummdidumm:
Thank you, but this is deliberate - it's meant to aid in solving the exercise. Therefore closing.

PR #161, opened 2022-12-16
regain focus when it gets lost by mistake

### LINKED ISSUE #150: Text input focus sometimes lost on rerender/error

Rich-Harris:
closes #150.

There are basically three ways that the editor could lose focus:

1. the user clicked on the page outside the editor
2. the user tabbed after hitting the `Ctrl-M` or `Ctrl-Shift-M` (on Mac) combo (which temporarily disables focus trapping, I learned today)
3. the iframe captures it

All we need to do, therefore, is determine whether a `focusout` event was caused by 1 or 2, and if so, recapture the focus. It seems to work reliably for me

### DIFF
--- a/src/routes/tutorial/[slug]/Editor.svelte
+++ b/src/routes/tutorial/[slug]/Editor.svelte
@@ -252,7 +252,7 @@
 
 				setTimeout(() => {
 					preserve_focus = true;
-				});
+				}, 0);
 			}
 		}}
 		on:focusin={() => {
--- a/src/routes/tutorial/[slug]/Editor.svelte
+++ b/src/routes/tutorial/[slug]/Editor.svelte
@@ -1,6 +1,5 @@
 <script>
 	import { dev } from '$app/environment';
-	import { monaco } from '$lib/client/monaco/monaco.js';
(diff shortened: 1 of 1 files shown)


src/routes/tutorial/[slug]/Editor.svelte
```diff
@@ -36,6 +35,8 @@
 	let w = 0;
 	let h = 0;
 
+	let preserve_focus = true;
```
Comment (dummdidumm) [inline src/routes/tutorial/[slug]/Editor.svelte:38]:
```suggestion
	/** 
	 * The iframe sometimes takes focus control in ways we can't prevent
	 * while the editor is focussed. Refocus the editor in these cases.
	 * This boolean tracks whether or not the editor should be refocused.
	 */
	let preserve_focus = true;
```

dummdidumm (approved):
(leaving a comment above the boolean so future selves know what the fuck is going on)

PR #162, opened 2022-12-16
correct Part 1 / Bindings / Select bindings code line

mattlehrer:
this could be a prettier / formatting issue? I am not sure if it's the same on all devices and platforms.


dummdidumm (approved):
Thank you! Yes, this is probably a result of formatting with Prettier and not updating the line number (aside: It's probably best to not refer to line numbers for this reason, this will get out of sync inevitably)

PR #165, opened 2022-12-20
Remove unnecessary and inconsistent quotation marks

Zokalyx:
# Reasons
- Quotation marks not necessary for code to compile
- The solution code does not use those quotation marks
- Creating strings inside of the quotation marks is uncomfortable


dummdidumm (approved):
Thank you!

PR #171, opened 2022-12-26
[fix] the position of single quote

tomoam:
The position of single quote and back quote was reversed, so I corrected it.


dummdidumm (approved):
Thank you!

PR #174, opened 2022-12-30
[fix] load images in Part 1 / Lifecycle / onMount excercise

ivanbtrujillo:
This PR is about to fix the images for the _Part 1 / Lifecycle / onMount_ excercise

While doing the tutorial, I noticed the images were not loading due to an error of type:
```
GET https://via.placeholder.com/150/61d552 net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep 200
```

By adding:

```
crossorigin="anonymous"
```
To the images, they appear as expected.

I've added the attribute to the initial and the solved state, and removed the warning from the Readme.

Without the attribute:
image

With the attribute:
image

### DIFF
--- a/content/tutorial/01-svelte/07-lifecycle/01-onmount/README.md
+++ b/content/tutorial/01-svelte/07-lifecycle/01-onmount/README.md
@@ -2,8 +2,6 @@
 title: onMount
 ---
 
-> The images in this exercise don't currently work. You can switch to the old tutorial instead: https://svelte.dev/tutorial/onmount
-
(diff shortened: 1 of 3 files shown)


dummdidumm (approved):
This is great, thank you!

Rich-Harris:
I'm afraid this still doesn't work for me — I think we need to revert it so that the message is still present.

I'm curious as to how you're seeing those images though — is that in dev, or preview? Or is it working for you in production on https://learn.svelte.dev/tutorial/onmount? If so, in which browser(s)?

PR #175, opened 2022-12-30
[fix] marked dependency for Part 1 / Bindings / Textarea inputs

ivanbtrujillo:
This PR fixes the error about Marked not loading. What I've done is add the Marked dependency at the excercise level in the same way Elizabot was added for _Part 1 / Lifecycle / beforeUpdate and afterUpdate_ excercise. 

node_modules folder is pushed to the repo (same way than Elizabot). 

Probably this is not the best way to do it, but it solves the issue for now. I think it worths to work on a script that runs npm install on any package.json under /content. This way we can remove the node_modules folder (by adding it to a .gitignore).

Before the fix:
image

After the fix:
image


content/tutorial/01-svelte/06-bindings/05-textarea-inputs/README.md
```diff
@@ -2,18 +2,16 @@
 title: Textarea inputs
 ---
 
-> This exercise doesn't currently work. You can switch to the old tutorial instead: https://svelte.dev/tutorial/textarea-inputs
-
 The `<textarea>` element behaves similarly to a text input in Svelte — use `bind:value`:
 
 ```svelte
-<textarea bind:value={value}></textarea>
+<textarea bind:value />
```
dummdidumm, line 8:
```suggestion
<textarea bind:value={value}></textarea>
```

content/tutorial/01-svelte/06-bindings/05-textarea-inputs/README.md
```diff
@@ -2,18 +2,16 @@
...
 ```
 
 In cases like these, where the names match, we can also use a shorthand form:
 
 ```svelte
-<textarea bind:value></textarea>
+<textarea bind:value />
```
dummdidumm, line 14:
```suggestion
<textarea bind:value></textarea>
```

dummdidumm (approved):
Thank you!

PR #176, opened 2022-12-30
[fix] beforeupdate and afterupdate excercise

ivanbtrujillo:
Eizabot dependency was not loading due to the following error:
```
Cannot find package '@sveltejs/adapter-auto' imported from /home/learnsveltedev-uthj/svelte.config.js
```
by just deleting the svelte.config.js file from that folder _content/tutorial/01-svelte/07-lifecycle/03-update/app-a_ it works.

Before the fix:
image

After the fix:
image

I've also updated the span style to make black the text received from the chatbot, because it has a white style on a lightgray background.


dummdidumm (approved):
Thank you!

PR #180, opened 2023-01-05
[fix] dice calculation in "GET handlers"

tillschander:
The tutorial on GET handlers includes a simulation of a dice roll:
`const number = Math.ceil(Math.random() * 6);`

While unlikely, the result of that dice roll might actually be 0. This is because Math.random() returns a number greater than or _equal to 0_.

This PR fixes that by using the follwoing code for the dice roll:
`const number = Math.floor(Math.random() * 6) + 1;`

----
Otherwise thanks for the great tutorial and congrats on the release of SvelteKit!


dummdidumm (approved):
Good catch, thank you!

PR #181, opened 2023-01-11
Implement backward and forward navigation options to iframed window

brandonmcconnell:
## Overview of changes

This PR adds backward and forward navigation options to iframed window. Because this sort of behavior isn't normally possible within the natural constraints of the History API since the history of the iframed window and that of its containing parent window operate together, I opted to manually track the history of the iframed window in two arrays— `history_bwd` and `history_fwd`.

#### Implementation/test cases & detailed steps for each

<details><summary>When any path is manually entered into the iframe's simulated address field… <em>(expand to see details)</em></summary><ul>
<li>the current <code>path</code> is appended to <code>history_bwd</code></li>
<li><code>history_fwd</code> is reset</li>
<li>the new <code>path</code> is naturally set to the entered path (current functionality)</li></ul></details>

<details><summary>When the "back" button is clicked… <em>(expand to see details)</em></summary><ul>
<li>the current <code>path</code> is prepended to <code>history_fwd</code></li>
<li>the last item in <code>history_bwd</code> is sliced off</li>
<li>the sliced item (the previous <code>path</code>) replaces the current <code>path</code></li></ul></details>

<details><summary>When the "forward" button is clicked… <em>(expand to see details)</em></summary><ul>
<li>the current <code>path</code> is appended to <code>history_bwd</code></li>
<li>the first item in <code>history_fwd</code> is sliced off</li>
<li>the sliced item (the next <code>path</code>) replaces the current <code>path</code></li></ul></details>

<details><summary>When any link within the iframe is clicked, it is picked up by <code>handle_message</code> which then falls back to using the <code>nav_to</code> function for handling and tracking the history accordingly, effectively… <em>(expand to see details)</em></summary><ul>
<li>the current <code>path</code> is appended to <code>history_bwd</code></li>
<li><code>history_fwd</code> is reset</li>
<li>the new <code>path</code> is naturally set to the entered path (current functionality)</li></ul></details>

## Notes

* The icons I used are custom ones I threw together in Illustrator, so feel free to use them without any risk of copyright violations, though I understand there are probably plenty of reasons to switch them, even just to stay on brand.
* I noticed after my initial successful tests that while the back/forward functionality worked as expected, the history persisted across tutorials, so I added an additional `reset_history` function that resets the history, which I trigger in the `try` clause within `afterNavigate`.

## Testing

This can be tested/demoed effectively using any of the routing examples:
* [Pages](https://learn-svelte-dfhs4unjp-svelte.vercel.app/tutorial/pages)
* [Layouts](https://learn-svelte-dfhs4unjp-svelte.vercel.app/tutorial/layouts)
* [Routing parameters](https://learn-svelte-dfhs4unjp-svelte.vercel.app/tutorial/params)

## Screen recording (UPDATED)

<details><summary>✅ Updated to include the flash-of-white fix implemented in commit <a href="https://github.com/sveltejs/learn.svelte.dev/pull/181/commits/644f5e3934fe4c9b34139c9cf74e8ad3f21f0360">644f5e3</a> (expand for details on fix)</summary>

<br />As of commit <a href="https://github.com/sveltejs/learn.svelte.dev/pull/181/commits/644f5e3934fe4c9b34139c9cf74e8ad3f21f0360">644f5e3</a>, the flash of white is now resolved by hiding the iframe (<code>display: none</code>, in addition to the existing <code>parentNode?.removeChild(iframe)</code> and <code>parentNode?.appendChild(iframe)</code>) until the iframe content has loaded, using an event listener set attached to <code>iframe</code> within the <code>onMount</code> and destroyed accordingly.</details>

https://user-images.githubusercontent.com/5913254/211909797-d4c5e8d4-8137-48d3-afff-b7bd3ead637b.mov

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.4
+lockfileVersion: 5.3
 
 specifiers:
   '@fontsource/roboto-mono': ^4.5.7
@@ -44,7 +44,7 @@ dependencies:
 devDependencies:
   '@sveltejs/adapter-auto': 1.0.0-next.90
   '@sveltejs/adapter-vercel': 1.0.0-next.85
-  '@sveltejs/kit': 1.0.0-next.586_svelte@3.54.0+vite@4.0.0
+  '@sveltejs/kit': 1.0.0-next.589_svelte@3.54.0+vite@4.0.0
   '@sveltejs/site-kit': 3.0.3
   '@types/diff': 5.0.2
   '@types/marked': 4.0.8
@@ -53,7 +53,7 @@ devDependencies:
   diff: 5.1.0
   esbuild: 0.14.54
   prettier: 2.8.1
-  prettier-plugin-svelte: 2.9.0_sro2v6ld777payjtkjtiuogcxi
+  prettier-plugin-svelte: 2.9.0_prettier@2.8.1+svelte@3.54.0
   svelte: 3.54.0
   svelte-check: 2.10.2_svelte@3.54.0
   tiny-glob: 0.2.9
@@ -356,8 +356,8 @@ packages:
       - supports-color
     dev: true
 
-  /@sveltejs/kit/1.0.0-next.586_svelte@3.54.0+vite@4.0.0:
-    resolution: {integrity: sha512-lTYWy4voh/r4jz8qnurIATyrH2ZgHPUswYQ9KauyASSjQGYeB1TPDgwcafM/LtgMxpcvFgWsx2KeflykSHXKxg==}
+  /@sveltejs/kit/1.0.0-next.589_svelte@3.54.0+vite@4.0.0:
+    resolution: {integrity: sha512-5ABRw46z9B+cCe/YWhcx+I/azNZg1NCDEkVJifZn8ToFoJ3a1eP0OexNIrvMEWpllMbNMPcJm2TC9tnz9oPfWQ==}
     engines: {node: '>=16.14'}
     hasBin: true
     requiresBuild: true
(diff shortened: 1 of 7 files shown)


Rich-Harris:
Sorry, I dropped the ball here, meant to review this PR sooner. I think we should revert it, so I've opened #187. The back/forward buttons in this PR cause the iframe to be reloaded, which gives a very misleading impression of how history navigation works in SvelteKit. It's better to not have history UI than to have UI that causes full page reloads

brandonmcconnell:
@Rich-Harris I might be wrong here, but I think the reload is caused by the iframe being removed and re-added to the screen, alongside the bwd/fwd history changes.

According to the comments surrounding that section…

```
// removing the iframe from the document allows us to
// change the src without adding a history entry, which
// would make back/forward traversal very annoying
```

…removing the iframe altogether and re-injecting it to the page is supposed to "change the src without adding a history entry" though I do not see that actually working in Chrome, even before my PR. Does that work in other browsers? As I navigate around one iframe, I still have to click the browser's back button repeatedly to traverse through every navigation from the iframe.

I can confirm that removing the `set_iframe_src` function and swapping its usage for `iframe.src =` resolves this issue, though again, I can't speak to whether any browsers were actually benefitting from `set_iframe_src`.

Let's go ahead and revert this PR for now, and either re-open this PR or I can create a new PR that builds on this one and addresses those concerns.

cc @dummdidumm

brandonmcconnell:
@Rich-Harris @dummdidumm One idea here which might remove much of the complexity around this is to sync the iframe history/URL state with the browser's so that as the iframe's URL changes, there could be URL param in the browser like `frame_src=${path}` so that clicking back in the browser or the iframe essentially trigger the same action and go back one step whether in the iframe or the browser depending on the user's actions. Of course, it would be better to just completely isolate the iframe's history apart from the browser's, but in all my testing and research, there doesn't appear to be a way to achieve that, including the "removing/adding the iframe node" hack.

I'm not sure if/how/where/when that work(ed), and I'm honestly not criticizing it; it may very well work in some places. It just hasn't worked in my testing, which granted was limited to Chrome.

PR #184, opened 2023-01-13
add comma

dnsbty:
The tutorial shows adding a new update method to the object being returned, but without the comma showing up there. This results in a syntax error if copied exactly, and it confused me for a minute.


dummdidumm (approved):
Thank you!

PR #187, opened 2023-01-23
Revert "Implement backward and forward navigation options to iframed window"

Rich-Harris:
Reverts sveltejs/learn.svelte.dev#181. Apologies, I should have been more on top of this


dummdidumm:
I opened #193 as an alternative. It keeps the navigation feature while fixing its flaws (and the navigation history pollution problem in general)

PR #189, opened 2023-01-23
chore: cleanup code

### LINKED ISSUE #139: make the code better
this isn't, like, an actionable issue or anything. i just get a bad feeling when I look at `/tutorial/[slug]/+page.svelte` that it's chock full of race conditions. needs to be simplified and robustified

dummdidumm:
closes #139

- move adapter stuff into its own component `Output.svelte`
- move icons into one folder
- move filetree into route where it's used and integrate more obviously with state
- make it more robust to rapid editing / resolve / switch actions through queue-like implementation in `adapter.js`
- make state easier to reason about by moving it into `state.js` with proper functions instead of writing to different writables all other the place


src/routes/tutorial/[slug]/filetree/File.svelte
```diff
@@ -1,16 +1,17 @@
...
 
 	/** @type {import('$lib/types').FileStub} */
 	export let file;
 
-	const { selected, endstate, select, rename, remove, readonly } = context.get();
+	const { rename, remove, readonly } = context.get();
 
 	/** @type {'idle' | 'renaming'} */
-	let state = 'idle';
+	let editing_state = 'idle';
```
Comment (Rich-Harris) [inline src/routes/tutorial/[slug]/filetree/File.svelte:12]: should we just make this a boolean?

src/routes/tutorial/[slug]/Output.svelte
```diff
@@ -0,0 +1,284 @@
...
+					console.error(e);
+				}
+
+				loading = false;
+			} else if (state.status === 'update' && state.last_updated) {
+				const reload = await adapter.update([state.last_updated]);
+				if (reload) {
+					schedule_iframe_reload();
+				}
+			}
```
Comment (Rich-Harris) [inline src/routes/tutorial/[slug]/Output.svelte:62]: this stuff feels racy — i feel like this logic belongs elsewhere, though i don't have a great design to offer right now

Comment (dummdidumm) [reply src/routes/tutorial/[slug]/Output.svelte:62]: The race conditions should be handled internally in `adapter.js`, though I now realize I'm not making use of the `"cancelled"` outcome here

Rich-Harris (approved):

PR #192, opened 2023-01-23
possible tweaks

Rich-Harris:
suggestions for #189 — calling `update` and `set` directly feels slightly nicer to me than having a `_state` variable floating around, and creating a new state object seems somehow nicer than mutating the existing one.

also, replaces `editing_state: 'idle' | 'remaining'` with `renaming: boolean`


dummdidumm (approved):
Makes sense!

PR #193, opened 2023-01-24
fix: better iframe navigation

dummdidumm:
alternative to #187 which keeps the back/fwd navigation while also fixing the weird history pollution from iframe to parent (which was there regardless of the back/fwd navigation feature)

- uses postMessage to trigger the inner SvelteKit's goto method
- monkeypatches iframe's history.pushState to use replaceState instead to not create history entries, keeping the parent window's navigation order intact

### DIFF
--- a/content/tutorial/common/src/__client.js
+++ b/content/tutorial/common/src/__client.js
@@ -15,53 +15,21 @@ window.alert = (message) => {
 };
 
 window.addEventListener('message', async (e) => {
-	if (e.data.type === 'fetch') {
-		const names = e.data.names;
-
-		const transformed = await Promise.all(
-			names.map(async (name) => {
-				const res = await fetch(name);
-				return {
-					name,
-					code: await res.text()
-				};
-			})
-		);
-
-		const css_files = [];
-
-		for (const { name, code } of transformed) {
-			if (
-				name.endsWith('.svelte') &&
-				code.includes('svelte&type=style&lang.css')
-			) {
-				css_files.push(name + '?svelte&type=style&lang.css');
-			}
-		}
-
-		if (css_files.length > 0) {
-			const css_transformed = await Promise.all(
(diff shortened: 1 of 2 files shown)


Rich-Harris:
This will break any tutorials that involve `navigation.type`, no?

Rich-Harris:
(I think we should undo #181 and work on this at leisure, rather than leaving the current UI in place any longer)

dummdidumm:
What do you mean by `navigation.type`?
(if you're nervous keeping the current state active we can do revert first if you want)

Rich-Harris:
In `beforeNavigate` and `afterNavigate` etc — `navigation.type` should be `popstate` but will be `link` https://kit.svelte.dev/docs/types#public-types-navigationtype. I'll do the revert, then we can figure this stuff out later

dummdidumm:
Ah I see. We could
- just not worry about this case in the respective tutorial
- specifically disable the pushState hack and use pop/push delta commands instead of `goto` for that tutorial
- disable the pushState hack and use pop/push delta for all tutorials (with the drawback that it pollutes the parent history like before)

I don't think that should hold us back though, since without this feature there wouldn't even be a way to trigger the `popstate` navigation type within the iframe.

brandonmcconnell:
@dummdidumm @Rich-Harris Another idea could be to allow the iframe state to work naturally until we see some movement on https://github.com/whatwg/html/issues/6501. That meaning, we could mesh the iframe history with the window's history and have window preserve the iframe state by storing the iframe's current path in a URL param, like…

```
https://learn.svelte.dev/tutorial/pages?path=/about
```

brandonmcconnell:
That way, it's not so much pollution as it is standard history tracking where there are no redundant polluted entries, but a bunch of unique entries that are all meaningful when they occur.

PR #197, opened 2023-01-27
Changed header text to title case

mouise1111:
Changed header text to title case for consistency in Todos component


Rich-Harris:
Not sure I follow — consistency with what? The subsequent chapters all use `todos`

PR #198, opened 2023-01-27
Separated label and input for Todos form

mouise1111:
Improved form readability and accessibility by separate label and input in Todos component


dummdidumm:
Thank you, but this change doesn't have any effect visual-wise and the current solution is more accessible, as you can also click on the label to focus the input - therefore closing.

PR #200, opened 2023-01-29
Text color updated to black

mouise1111:
text color updated to black since input box is white, visibility issue fixed


dummdidumm (approved):
Thank you!

PR #207, opened 2023-02-10
fix: restart Vite if tutorial has a .env file

### LINKED ISSUE #206: Errors occur when switching from a tutorial without .env to a one with .env
There are two problems related to `.env`.

The first is that if you switch from a tutorial without `.env` to one with `.env` and wait 10 seconds or so, you get an error.

https://user-images.githubusercontent.com/29677552/217544265-e8b4ee46-8b67-4002-b369-ee93fa3f48ab.mov

Second, when switching from a tutorial without `.env` to one with `.env`, environment variables loaded from `.env` cannot be used (the first problem occurs after 10 seconds of switching, so you must operate quickly to reproduce it ).

https://user-images.githubusercontent.com/29677552/217544350-e6acc9dd-fa1f-4a83-8fd8-c3f7767db454.mov

Neither problem occurs when directly opening a tutorial with `.env` (or when switching from a tutorial with `.env`).

The first problem is because that learn.svelte.dev expects Vite to restart if the tutorial has `.env` but Vite only restarts if the `.env` file is changed, not added or removed.

I will try to find out about the second problem as well.

tomoam:
fixes #206 

There are two problems with #206.

For the first problem, as expected, learn.svelte.dev waits for Vite dev server to restart when a `.dev` file is also added, but Vite only restarts when the `.env` file is changed.

- Vite

https://github.com/vitejs/vite/blob/d953536aae448e2bea0f3a7cb3c0062b16d45597/packages/vite/src/node/server/index.ts#L486-L503

https://github.com/vitejs/vite/blob/d953536aae448e2bea0f3a7cb3c0062b16d45597/packages/vite/src/node/server/hmr.ts#L54-L72

- learn.svelte.dev

https://github.com/sveltejs/learn.svelte.dev/blob/1c3bebf6dd07596924066bf43edd20082700ce08/src/lib/client/adapters/webcontainer/index.js#L235-L243

So, you might assume that just make learn.svelte.dev not wait for Vite dev server to restart when a `.env` file is added, but that would not solve the second problem.

As for the second problem, it is due to Vite's behavior. That is, when a `.env` file is added, the Vite dev server must be restarted in order to make the environment variables loaded from the `.env` file available.

In other words, if switching to a tutorial with a `. env`  file, the Vite dev server must be restarted. However, I looked through the Vite documentation and could not find a way to restart Vite.

This PR fix is hacky, but it solves both problems. Of course, if you have a better idea, I would like to do so.

### DIFF
--- a/src/lib/client/adapters/webcontainer/index.js
+++ b/src/lib/client/adapters/webcontainer/index.js
@@ -183,6 +183,7 @@ export async function create(stubs, callback) {
 			// are not available until Vite is restarted. By creating a dummy `.env` file, it will
 			// be recognized as changed when the real `.env` file is loaded into the Webcontainer.
 			// This will invoke a restart of Vite. Hacky but it works.
+			// TODO: remove when https://github.com/vitejs/vite/issues/12127 is closed
 			if (!previous_env && current_stubs.has('/.env')) {
 				await vm.run({ command: 'touch', args: ['.env']});
 			}
--- a/src/lib/client/adapters/webcontainer/index.js
+++ b/src/lib/client/adapters/webcontainer/index.js
@@ -119,11 +119,6 @@ export async function create(stubs, callback) {
 			/** @type {import('$lib/types').Stub[]} */
 			const to_write = [];
 
-			/** @type {import('$lib/types').FileStub=} */
-			let previous_env;
-			/** @type {import('$lib/types').FileStub=} */
-			let new_env;
-
 			for (const stub of stubs) {
 				if (stub.type === 'file') {
 					const current = /** @type {import('$lib/types').FileStub} */ (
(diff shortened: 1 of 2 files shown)


tomoam:
Alternatives considered

- Add an blank `.env` file to the `content/tutorial/common` directory.
	- This will treat a `.env` files added as changed. This way we don't have to modify the code, but a `.env` file would appear in the SvelteKit tutorials, which may confuse the user who is trying to learn.

- Restart Vite dev server programatically.
	-  When Switching from a tutorial without a `.env` to one with it, kill Vite dev server process and then restart Vite dev server with `turbo run dev` (see [implementation example](https://github.com/sveltejs/learn.svelte.dev/commit/648cf7c9f75d8da656a9a52a96a87ec9c4ceaac9)). This way seems a royal road, but has two problems: first, it is slow to start a tutorial when switching. and second, it relies heavily on the Webcontainer API, which is still in alpha.

Rich-Harris (approved):

Rich-Harris:
Thank you for the PR — brilliant sleuthing! I've opened an issue on Vite, hopefully we can remove this hack one day: https://github.com/vitejs/vite/issues/12127

PR #209, opened 2023-02-10
[fix] remove redundant spaces

sadnesszephyr:
Removed redundant whitespaces to make it consistent with `app-a` and docs code style in general


dummdidumm (approved):
Thank you!

PR #216, opened 2023-02-16
Replace XSS abbreviation with Cross-Site Scripting and link to OWASP

orrefailaT:
Mainly for the benefit of less experienced programmers that may not be familiar with XSS vulnerabilites, I think replacing the abbreviation and linking to further information would be useful in driving this point home.


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #218, opened 2023-02-18
fix: Correct paths in layout-data tutorial

grschafer:
Paths are missing `routes` folder (`src/blog/...` should instead be `src/routes/blog/...`)


Rich-Harris (approved):

Rich-Harris:
thank you!

PR #220, opened 2023-02-21
webcontainer v1

Rich-Harris:


Conduitry:
This appears to still be something that opens up an external iframe. Are there any alternatives to that yet? Is that on StackBlitz's roadmap?

PR #222, opened 2023-02-21
fix: setTimeout for time-consuming test

tomoam:
It seems that CI is failing because one of the tests I added is flaky, sorry.
By setting a longer timeout value for time-consuming tests, this problem can be alleviated.


Rich-Harris:
thank you!

PR #224, opened 2023-02-21
Update README.md

Apsis:
Proposed change to show <input type="number"> and <input type=number> both function!


dummdidumm:
Thank you, but I think it doesn't fit into the given tutorial. It's a concept that applies to attribute values on HTML elements in general. Therefore closing.

PR #228, opened 2023-02-21
make mapbox tutorial work

### LINKED ISSUE #124: Context Mapbox exercise broken
https://learn.svelte.dev/tutorial/context-api is broken, because `mapbox-gl` doesn't exist. Probably solvable by adding it to the bundle. In general, this shows that we're missing the unpkg-load-any-npm-library-on-the-fly feature that is working in the regular Svelte REPL, which is bad.

Rich-Harris:
this fixes #124... at a cost. It has a sprawling dependency graph so we end up adding a _lot_ of stuff to `node_modules`, which has to be downloaded. Maybe we can bundle it better, or maybe installing it with `turbo` (the WebContainer `npm` stand-in) would work. Will investigate further


Conduitry:
Holy shit that's a diff. Why are we checking in all these dependencies? And why do we seem to have a bunch of mapbox tests? Are they part of their npm package?

Rich-Harris:
Yeah. It's not packaged very well

Rich-Harris:
> Why are we checking in all these dependencies?

We _could_ avoid checking `node_modules` in but I think this is one of those cases where forcing the cost of packages to be visible really does make sense, since all these files will be downloaded when the webcontainer starts up

Conduitry:
Regardless of how it's packaged, something still feels very wrong about checking these all into the repo. I see we're only gitignoring `/node_modules` rather than `node_modules` at all directory levels. Is there a reason for this? Can we at least pull mapbox in at build time rather than having it all live vendored in this repo?

dummdidumm:
Good idea, some predev/build script that downloads a fixed version into the respective folders would be great

Rich-Harris:
> Is there a reason for this?

As I say, forcing it to be visible is arguably a good thing in this context. We wouldn't be having this conversation if it hadn't been done that way! I'd much rather we find a way to bundle `mapbox-gl` ourselves than have the user download 100Mb of `node_modules` that gets generated at build time.

tomoam:
Personally, I would be against committing minified `.js` files to this repository. Minified `.js` files are less readable, and it is harder to detect if there is malicious code in it. If it is just this one time, and if it is done by the maintainer team, I don't see a problem. But in the future, if someone with malicious intent is about to commit a minified `.js` files containing malicious code when adding a new tutorial, it would be hard to detect it.

So I think it would be better to write a script that pre-builds and copies the code, and run it at the same time of running `create-common-bundle`.

In the case of this tutorial, it seems to work with only `dist/mapbox-gl.js` among the dependencies. I've prepared a example branch and a preview environment in the forked repository. If you find them useful I will submit a PR, otherwise please disregard.

- https://github.com/svelte-jp/learn.svelte.dev/pull/25/files
- https://learn-svelte-micnmfmsd-tomoam.vercel.app/tutorial/context-api

Rich-Harris:
going to close this since the mapbox example has been replaced with George Nees https://learn.svelte.dev/tutorial/context-api

PR #235, opened 2023-02-26
Clarify that you have only finished the Svelte section of the tutoria…

b1ue64:
…l, not the entire thing.


b1ue64:
actually, it's probably more fitting to move it to the end of Advanced SvelteKit

PR #237, opened 2023-03-07
Update README.md

ashasteria:
The file edited should be +page.server.js, not +layout.server.js, according to the interactive editor.


dummdidumm:
The sentence is correct as is - it's talking about the general case, not the specific code in the tutorial. Therefore closing - thank you.

PR #241, opened 2023-03-09
Remove irrelevant change from tutorial

lukasnys:
In `01-svelte/07-lifecycle/02-ondestroy` the boolean `open` switches value from `false` to `true`.

While solving the tutorial, I had to use the "solve" functionality in order to figure out what I was missing. This change of value seems irrelevant to the lesson itself and might bring confusion to others.


Rich-Harris (approved):

Rich-Harris:
thanks!

PR #245, opened 2023-03-10
fix: prevent iframe from reloading when a folder is clicked

tomoam:
Immediately after opening learn.svelte.dev, if you click on the folder, the iframe will reload.

https://user-images.githubusercontent.com/29677552/224252481-a942d749-54c5-4061-98a2-db73268b74dc.mov

Note that I added `expand` to `status` for this fix, but it may not be the proper term (because I couldn't think of a better term with my poor English, sorry...).


dummdidumm (approved):
Thank you!

Issue #248, opened 2023-03-10
WebContainers + HMR = pain


Rich-Harris:
The issue is that we're creating the adapter in a lifecycle hook when really it should be a persistent singleton, like `state`. I also think we could simplify the code a lot by treating it as such

Rich-Harris:
closed via #249

PR #249, opened 2023-03-10
overhaul adapter code

Rich-Harris:
This simplifies the adapter code per #248. There was a bit too much indirection previously (also we were conflating state and events, which I think is a bit of a code smell) and I found it quite hard to make changes — there's still room for improvement but I'm much happier with this. 

There are a couple of glitches to work through still:

* [ ] rapidly navigating between e.g. [/tutorial/universal-load-functions](https://learn-svelte-dev-git-gh-248-svelte.vercel.app/tutorial/universal-load-functions) and [/tutorial/the-form-element](https://learn-svelte-dev-git-gh-248-svelte.vercel.app/tutorial/the-form-element) will result in a fallback error page, which sometimes doesn't self-repair
* [x] if you navigate to a path like `/about` on [/tutorial/pages](https://learn-svelte-dev-git-gh-248-svelte.vercel.app/tutorial/pages), that path will persist when you navigate to other exercises

### DIFF
--- a/src/routes/tutorial/[slug]/adapter.js
+++ b/src/routes/tutorial/[slug]/adapter.js
@@ -1,3 +1,17 @@
+import { writable } from 'svelte/store';
+import { browser } from '$app/environment';
+
+export const progress = writable({
+	value: 0,
+	text: 'initialising'
+});
+
+/** @type {import('$lib/types').Adapter} */
+// @ts-expect-error
+export const adapter = browser
+	? create_adapter((value, text) => { progress.set({ value, text }); })
+	: undefined;
+
 /**
  * @param {(progress: number, status: string) => void} callback
  * @returns {import('$lib/types').Adapter}
@@ -46,7 +60,6 @@ export function create_adapter(callback) {
 	current = init().then(() => true);
 
 	return {
-		init: current.then(() => {}),
 		get base() {
 			return adapter_base;
 		},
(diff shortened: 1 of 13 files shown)


src/routes/tutorial/[slug]/adapter.js
```diff
@@ -1,77 +1,88 @@
...
-		adapter_base = (await adapter_promise).base;
-	}
+export function subscribe(event, callback) {
+	subscriptions.get(event)?.add(callback);
 
-	// Keep track of what's currently running, and what's next
-	/** @type {Promise<boolean>} */
-	let current;
-	let token = {};
-	async function next() {
```
Comment (dummdidumm) [inline src/routes/tutorial/[slug]/adapter.js:27]: The simplification inside `Output.svelte` is great, but I think we need to keep this queue implementation. The new code is - I think - prone to race conditions when typing fast, toggling between solution and initial state fast etc.

Comment (Rich-Harris) [reply src/routes/tutorial/[slug]/adapter.js:27]: I haven't observed that to be the case, but if it is then I think we want to implement the queuing logic in `webcontainer/index.js`, where there's already an `await running` (which appears to only be used in the `reset` case) — having multiple layers was part of the reason the code was hard to work with. As far as `adapter.js` goes it should be fire-and-forget, I think

Rich-Harris:
still occasionally encountering errors when navigating between apps, but no more than currently, so i'll merge this to unblock other work

PR #252, opened 2023-03-13
switch to codemirror

Rich-Harris:
still various things to do

* [x] styling
* [x] focus management
* [x] see if there's any obvious functionality we're missing

once it works, it should be easier to add things like annotations, mobile support, etc


Rich-Harris:
The preview is noticeably faster to load (which is unsurprising if you look at the network tab). Monaco is a beast.

Rich-Harris:
(also, the console doesn't become unusable with bullshit errors/warnings)

Issue #47, opened 2022-06-08
Switch to CodeMirror


Rich-Harris:
Filing this under 'one day' — it would be great to use CodeMirror 6 instead of Monaco, since it works on mobile and is generally (I believe) more accessible. It would be a big lift though.

cassepipe:
I know CodeMirror supports a vim mode. It'd be great if it was exposed to the user.

Rich-Harris:
Switched to CodeMirror in #252. POC Vim mode in #253, but it needs some UI or something, and I don't have any great ideas off the top of my head

Rich-Harris:
vim mode can now be enabled by appending `?vim=true` to the URL

cassepipe:
Thanks for making this happen !

PR #253, opened 2023-03-14
vim mode

Rich-Harris:
https://github.com/sveltejs/learn.svelte.dev/issues/47#issuecomment-1445345202. Just leaving this here for now — it can't be merged like this, it needs to be hidden behind an option, but I'm not sure where such a thing should go


Rich-Harris:
Rather than add UI for this, I reckon we can get away with `?vim=true` for now. The setting persists to `localStorage`

Issue #12, opened 2022-05-09
Mobile layout


Rich-Harris:

dummdidumm:
A version of this is almost done, following the three-way-toggle design of the Svelte tutorial. The things that's left to do is cleaning up the code view, because the file tree takes up too much of the screen. It probably needs some kind of show/hide file tree mechanism.

Rich-Harris:
closed via #259

PR #254, opened 2023-03-14
mobile layout

Rich-Harris:
#12. The 'input' screen needs some work


Rich-Harris:
this isn't 'finished', but it works well enough that i'm going to merge it as-is then think about a redesigned version. i think it makes sense to have two screens rather than three — tutorial and editor/preview — with the file tree hidden behind a menu.

this will need some changes to the `<SplitPane>` component to avoid a flash of incorrect layout

PR #269, opened 2023-03-17
docs: add svelte:element/document chapters

dummdidumm:
from old tutorial - also updates some existing content


dummdidumm:
Ok just copying over the whole thing turns out to be more cumbersome than surgically updating and moving the folders

PR #270, opened 2023-03-17
docs: add svelte:element/document chapters

dummdidumm:
from old tutorial - also updates some existing content


dummdidumm:
I don't understand why I'm getting the "invalid special tag `svelte:document`" error on https://learn-svelte-nudpkxpd7-svelte.vercel.app/tutorial/svelte-document when clicking "solve" - I bumped Svelte, why doesn't it recognize that?

Issue #266, opened 2023-03-15
Failure to recover from certain error states


Rich-Harris:
If you follow the instructions on e.g. https://learn.svelte.dev/tutorial/component-events, the screen will go blank and it won't recover unless you hit the reload icon

Rich-Harris:
i think we can close this now

PR #272, opened 2023-03-17
queue all updates

Rich-Harris:
#266 seems to be caused by a race condition. this appears to fix it


Rich-Harris:
this PR pairs well with https://github.com/sveltejs/kit/pull/9441

PR #302, opened 2023-03-27
fix: prevent content from being reset

### LINKED ISSUE #298: Content is reset at unexpected times
I'm sure there's a simpler repro than this, but I'm making a note of what I observed just now for further investigation later.

1. Go to https://learn.svelte.dev/tutorial/textarea-inputs
2. Hit 'solve'
3. Change `- lists` to `- [links](https://example.com)`
4. Click the newly-rendered link in the `output` section
5. Close the tab

The content will have been reset (this happens for me in Firefox and Chrome).

Weirdly, you have to actually close the tab — if you leave it open and navigate back to the tutorial, the content won't have been reset. And if you drag the newly opened tab into a separate window and _then_ close it, nothing will happen. It's... strange.

tomoam:
Fixes #298 

After much trial and error, I think this fix is appropriate.

### DIFF
--- a/src/routes/tutorial/[slug]/Output.svelte
+++ b/src/routes/tutorial/[slug]/Output.svelte
@@ -41,6 +41,23 @@
 	/** @type {any} */
 	let timeout;
 
+	function reset_timeout() {
+		loading = false;
+
+		clearTimeout(timeout);
+
+		if (document.visibilityState === 'hidden') return;
+
+		timeout = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


tomoam:
Maybe if we update the SvelteKit version to 1.14 (thanks to sveltejs/kit#9497), we can remove some code (like `timeout`, `ping-pause`, etc) and this PR will no longer be needed.

PR #304, opened 2023-03-29
fix: codemirror scroll behaviour

### LINKED ISSUE #299: CodeMirror scroll behaviour is weird
When the editor has a lot of content, certain edits will cause the scroll to jump down

tomoam:
Fixes #299 

According to the codemirror documentation, [`setState`](https://codemirror.net/docs/ref/#view.EditorView.setState) reset the view, and may causes some side effects .

Currently `setState` is called every time the code is changed (not just when switching files). It can be fixed by calling it only when switching files

### DIFF
--- a/src/routes/tutorial/[slug]/Editor.svelte
+++ b/src/routes/tutorial/[slug]/Editor.svelte
@@ -51,8 +51,9 @@
 
 	$: reset($files);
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
nice, this is such an elegant fix 🙌

PR #303, opened 2023-03-28
chore: update @svelte/kit to v1.14.0

tomoam:
ref https://github.com/sveltejs/learn.svelte.dev/pull/302#issuecomment-1486584294

Thanks to sveltejs/kit#9497, it would be very beneficial for learn.svelte.dev to update the SvelteKit version to 1.14.

PR #309, opened 2023-04-03
fix: prevent content from being reset

### LINKED ISSUE #298: Content is reset at unexpected times

tomoam:
ref: https://github.com/sveltejs/learn.svelte.dev/pull/302#issuecomment-1486584294 , https://github.com/sveltejs/learn.svelte.dev/pull/303#issue-1643639716

Fixes #298 

SvelteKit's version has been updated to 1.14 (now 1.15), which improves recovery from errors. Now, code such as `ping` and `timeout` (the cause of #298, see [here](https://github.com/sveltejs/learn.svelte.dev/issues/298#issuecomment-1482954163)) should no longer be needed.

### DIFF
--- a/content/tutorial/common/src/__client.js
+++ b/content/tutorial/common/src/__client.js
@@ -1,19 +1,3 @@
-// Hack into the alert that's used in some tutorials and send a message prior to the alert,
-// else the parent thinks we lost contact and wrongfully reloads the page.
-// The drawback is that alert is no longer blocking, but no tutorial relies on this.
-const alert = window.alert;
-window.alert = (message) => {
-	parent.postMessage(
-		{
-			type: 'ping-pause'
-		},
-		'*'
-	);
-	setTimeout(() => {
-		alert(message);
-	});
-};
-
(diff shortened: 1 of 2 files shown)


Rich-Harris:
I have a _slight_ bit of nervousness about this (it feels like there could be some unanticipated way to get into a broken state) but this is way nicer, so let's give it a whirl and see if anything breaks. thanks!

PR #310, opened 2023-04-03
fix: prevent build failure due to cache files

tomoam:
The following are logs of a build failed in Vercel (these are from folk site for Japanese, but probably the same error occurs in learn.svelte.dev).

```log
23:29:08.034 | Error: ENOENT: no such file or directory, open 'content/tutorial/01-svelte/06-bindings/05-textarea-inputs/README.md'
23:29:08.034 | at Object.openSync (node:fs:600:3)
23:29:08.034 | at Object.readFileSync (node:fs:468:35)
...
```

This is due to old files still in Vercel.

### DIFF
--- a/scripts/create-common-bundle/index.js
+++ b/scripts/create-common-bundle/index.js
@@ -5,6 +5,10 @@ import glob from 'tiny-glob/sync.js';
 import { fileURLToPath } from 'url';
 import { execSync } from 'child_process';
 
+if (!!process.env.VERCEL) {
+	execSync('git clean -d -f content/tutorial');
+}
+
 const cwd = 'content/tutorial/common';
 
 execSync('npm ci', { cwd });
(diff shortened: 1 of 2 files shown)


scripts/create-common-bundle/index.js
```diff
@@ -5,6 +5,10 @@ import glob from 'tiny-glob/sync.js';
 import { fileURLToPath } from 'url';
 import { execSync } from 'child_process';
 
+if (!!process.env.VERCEL) {
+	execSync('git clean -d -f content/tutorial');
+}
```
tomoam, line 10: When moving or editing exercises locally, we don't want to accidentally remove files. For this reason, this should only be run on Vercel.

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #319, opened 2023-04-05
fix: refresh address bar

### LINKED ISSUE #318: Address bar no longer updates
#309 seems to have caused the Address bar to no longer be updated.
Probably I can solve it using MutationObserver instead of reverting #309.

https://user-images.githubusercontent.com/29677552/229971837-1be841a9-8ccc-4ca0-b8c0-aaac76fd42cd.mov

tomoam:
Fixes #318


Rich-Harris (approved):

Rich-Harris:
love the fact that the URL bar now updates instantly

PR #320, opened 2023-04-05
Fix user comment bug

monooso:
User comments appear as empty chat bubbles because we clear the input value _before_ recording the comment string. This fixes that.


Rich-Harris:
ah whoops, i broke this recently. thank you!

PR #326, opened 2023-04-06
nicer action exercise

Rich-Harris:
This adds a more engaging and understandable exercise for actions.

Unfortunately it reveals a bug — it's currently impossible to complete the exercise, because the iframe keeps snatching the focus. I'm not sure if that's something that broke recently or if this was always broken, but it would be good to fix it if possible.

If not, the workaround would be to not show the menu initially, but that would be sad.

### DIFF
--- a/content/tutorial/01-svelte/04-logic/04-each-blocks/app-a/src/lib/App.svelte
+++ b/content/tutorial/01-svelte/04-logic/04-each-blocks/app-a/src/lib/App.svelte
@@ -7,21 +7,21 @@
 
 <div>
 	<button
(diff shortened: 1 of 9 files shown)


Rich-Harris:
Okay I think I figured it out — the `iframe_took_focus` event is now fired for all `focusin` events inside the iframe unless they're the result of user action, using the following heuristic:

* if the `focusin` event happened while a pointer or key was down, it was user initiated
* if the target is `<body>` and there's a `relatedTarget`, it was user initiated (on hydration, `<body>` will be focused, but there'll be no `relatedTarget`)

In all other cases, the editor should regain focus if the iframe steals it.

PR #327, opened 2023-04-06
Tippy

Rich-Harris:
WIP — a better exercise for actions with parameters


dummdidumm:
I don't like that this adds a dependency (because dependency management is super awkward for learn.svelte.dev 😄 )

Rich-Harris:
It shouldn't cause any ongoing pain though. And if it does, we should come up with a solution to the pain, rather than avoiding dependencies

PR #328, opened 2023-04-06
fix: Rest parameters exercise

### LINKED ISSUE #321: "Rest parameters" exercise has a blank display
https://learn.svelte.dev/tutorial/rest-params

This is because `content/tutorial/common/src/routes/+page.svelte` exists.

image

tomoam:
Fixes #321 , by enabling `__delete` to be used in `app-a` as well as `app-b`.


Rich-Harris:
Oops, I fixed this separately in 81eb254dab39f4e7290d56ad9c54a9942d83b95b without realising this PR existed!

It turns out the approach in this PR doesn't fully solve the problem, because `src/routes/+page.svelte` is part of `content/tutorial/common` — if you go to https://learn-svelte-dev-git-fork-tomoam-fixes-321-svelte.vercel.app/tutorial/rest-params and visit `/` in the iframe, you get the blank page (instead of a 404) even though that file isn't considered part of the exercise. We need to actively tell the WebContainer to delete the file rather than just not including it. I'll close this since it was fixed elsewhere

Issue #313, opened 2023-04-03
Pop-out window


Rich-Harris:
Very occasionally it's nice to be able to view the output in a separate window/tab. Might be cool to have a button to open one, rather than digging around in the DOM to find the URL

tomoam:
I very much agree.
I think this could be implemented by adding something like:

```svelte
<a href={$base + path} target="_blank">Open in New Tab</a>
```
...but would need to be carefully tested for errors, etc.

I will create a draft PR to share and discuss the idea.

Rich-Harris:
closed via #329

PR #329, opened 2023-04-06
feat: open in new tab

tomoam:
ref #313 

WebContainer is amazing.

In the meantime, it seems to work, but I don't know what errors to expect yet.
I also think there are some things to consider, like where to put the `a` element, what to do with the icons, etc.

https://user-images.githubusercontent.com/29677552/230345914-5d13de03-b25a-4e33-98b9-85545ba65ec8.mov

https://user-images.githubusercontent.com/29677552/230345974-e68ee843-36a9-46b2-b71a-2f8d6240834f.mov

### DIFF
--- a/src/routes/tutorial/[slug]/Chrome.svelte
+++ b/src/routes/tutorial/[slug]/Chrome.svelte
@@ -9,6 +9,9 @@
 	/** @type {boolean} */
 	export let loading;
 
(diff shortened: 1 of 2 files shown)


Rich-Harris:
I added an icon in #330 which is now merged, so I'll close this — thanks!

PR #346, opened 2023-04-08
fix: accept autocompletions with Tab key without losing focus

### LINKED ISSUE #324: losing focus when accepting autocompletions
Not every time, but sometimes:

https://user-images.githubusercontent.com/1162160/230167246-5dbff2f7-994c-42c1-b178-bfb48b012dc0.mov

tomoam:
Fixes #324 

The code removed in this PR worked in combination with the processing that was previously in `on:focusout`.
It should no longer be needed now.

As a side note, while I was testing this PR, I learned that Monaco and Codemirror have different tab trapping keys.
https://codemirror.net/examples/tab/

### DIFF
--- a/src/routes/tutorial/[slug]/Editor.svelte
+++ b/src/routes/tutorial/[slug]/Editor.svelte
@@ -206,15 +206,6 @@
 <div
 	class="container"
 	bind:this={container}
-	on:keydown={(e) => {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
i love PRs that are just deleting code! thank you

PR #350, opened 2023-04-10
Remove extraneous call to loop function

nicholas-wilcox:
This extra call to the `loop` function isn't present in `app-b/src/lib/App.svelte`, so it should be removed from the text of the tutorial.

I made this commit and PR completely through GitHub, so that's too blame for the change on the last line. I don't know if there's a "newline at EOF" policy for this repository.


Rich-Harris:
thank you!

PR #352, opened 2023-04-10
Streaming

Rich-Harris:
This is a draft PR, because I'm not sure this is possible with WebContainers. For whatever reason, streaming isn't working here


Rich-Harris:
Specifically, if you visit https://learn-svelte-dev-git-streaming-svelte.vercel.app/tutorial/streaming and hit 'solve', the page buffers, making navigation very slow

Rich-Harris:
> the page buffers, making navigation very slow

note that you need to hit the refresh button to see this (it streams correctly on soft navigation, but it buffers on hard navigation)

the content here is out of date, needs to be updated for SK2

PR #353, opened 2023-04-11
remove "here be dragons" warning

Rich-Harris:
the tutorial is... basically done? there's still a bunch of issues open but they're non-critical, we can fix them at leisure i think.

similarly, there's some more content we _could_ add, but the tutorial in its current form is pretty comprehensive — certainly comprehensive enough that we should probably start redirecting here from https://svelte.dev/tutorial.

draft because it would be nice to adopt `site-kit@4` (see e.g. https://github.com/sveltejs/site-kit/pull/88)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
 		"@playwright/test": "^1.31.2",
 		"@sveltejs/adapter-vercel": "2.3.1",
 		"@sveltejs/kit": "^1.15.0",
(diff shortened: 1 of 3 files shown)


benmccann (approved):
ship it!

dummdidumm:
Nice! I think we can't just redirect from the old one yet since not all browser support web containers - but we certainly can add a banner or sth like that to advertise it there.

Issue #348, opened 2023-04-10
Failure to recover from broken `handle` hook


Rich-Harris:
Following the instructions on https://learn.svelte.dev/tutorial/handle (which I just added) causes the page to get into a broken state.

At the end, when clicking on the link to `/ping`, the URL bar doesn't update (perhaps we can do something sneaky like intercepting link clicks and posting an early `ping` message?). Focusing the input and hitting `enter` doesn't cause the page to reload.

PR #355, opened 2023-04-11
fix: reload with Enter key while in URL input

tomoam:
Fixes part of #348

### DIFF
--- a/src/routes/tutorial/[slug]/Chrome.svelte
+++ b/src/routes/tutorial/[slug]/Chrome.svelte
@@ -23,6 +23,12 @@
 		on:change={(e) => {
 			dispatch('change', { value: e.currentTarget.value });
 		}}
+		on:keydown={(e) => {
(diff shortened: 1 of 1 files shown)


src/routes/tutorial/[slug]/Chrome.svelte
```diff
@@ -23,6 +23,12 @@
 		on:change={(e) => {
 			dispatch('change', { value: e.currentTarget.value });
 		}}
+		on:keydown={(e) => {
+			if (e.key === 'Enter') {
+				dispatch('change', { value: e.currentTarget.value });
+				e.currentTarget.blur();
```
Comment (tomoam) [inline src/routes/tutorial/[slug]/Chrome.svelte:29]: Without this, the focus will remain on the input.
It should focus on the iframe, but I have no idea of a better way to do it, so I did it this way.

Rich-Harris:
thanks! this change required a small additional fix, which i added in #358 — closing this in favour of that PR

Issue #341, opened 2023-04-08
Race condition when restarting server


Rich-Harris:
Not quite sure what's going on here, but rapidly editing an `.env` file seems to cause something to get stuck:

https://learn.svelte.dev/tutorial/env-static-public

https://user-images.githubusercontent.com/1162160/230727682-285d9739-7292-42d8-858a-fe20a5a55ddb.mov

tomoam:
Vite restarts on `.env` file update. 
https://github.com/vitejs/vite/blob/09350c69ada6a43b26241b99a0d4f8aa4283ade2/packages/vite/src/node/server/hmr.ts#L44-L75

So we might have to wait for `server-ready` when updating the `.env` file as well.

tomoam:
I agree completely.
I tried to implement waiting for `server-ready` in the `update` method when updating `.env` and found out why that mechanism is needed.

tomoam:
Related to this, in rare cases, there seems to be an error when moving from an exercise with `.env` file to one without `.env` file.


https://user-images.githubusercontent.com/29677552/231753951-49ac5ad7-7b42-4565-b75f-3ac933e796d6.mov

PR #364, opened 2023-04-13
chore: add scheduler

tomoam:
Fixes https://github.com/sveltejs/learn.svelte.dev/issues/341#issuecomment-1501739958

This PR includes only the mechanism shown in https://github.com/sveltejs/learn.svelte.dev/issues/341#issuecomment-1501739958.
If you would prefer to include this in #363, please let me know and I will do so.


tomoam:
I have found a more efficient solution than this PR. Since it is the same file to be changed as #363, I close this PR and will commit to #363.

PR #372, opened 2023-04-16
Fix file name referenced in README

AhmadMayo:


tomoam:
Thank you, but it seems to be a duplicate of #369.

PR #374, opened 2023-04-24
fix filename typo in use enhance lesson

caseyg:
noticed a tiny typo where the tutorial refers to `/// file: App.svelte` instead of `src/routes/+page.svelte`


Rich-Harris:
looks like this was fixed in another PR: #380. thank you!

PR #376, opened 2023-05-02
fix: correct display for warnings

### LINKED ISSUE #322: Race condition around compiler warnings
Hard to repro completely reliably, but there's some race condition that allows out of date compiler warnings to be shown. This issue is mostly a placeholder until I have time to look into it more closely

tomoam:
fixes #322

### DIFF
--- a/src/lib/client/adapters/webcontainer/index.js
+++ b/src/lib/client/adapters/webcontainer/index.js
@@ -4,6 +4,12 @@ import AnsiToHtml from 'ansi-to-html';
 import * as yootils from 'yootils';
 import { escape_html, get_depth } from '../../../utils.js';
 import { ready } from '../common/index.js';
+import { warnings } from '../../../../routes/tutorial/[slug]/state.js';
+
+// TODO state.js and adapter.js should be moved from `routes/tutorial/[slug]` to `lib/client`?
+/**
+ * @typedef {import("../../../../routes/tutorial/[slug]/state.js").CompilerWarning} CompilerWarning
+ */
 
 const converter = new AnsiToHtml({
 	fg: 'var(--sk-text-3)'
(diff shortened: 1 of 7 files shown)


src/lib/client/adapters/webcontainer/index.js
```diff
@@ -45,12 +51,40 @@ export async function create(base, error, progress, logs) {
 		}
 	});
 
+	/** @type {Record<string, CompilerWarning[]>} */
+	let $warnings;
+	warnings.subscribe((value) => $warnings = value);
```
dummdidumm, line 56: With this there's a dependency loop between the adapter and the output. Can we move the `warnings` store into `tutorial/[slug]/adapter.js` (like the other stores) to decouple this?

tomoam: Thanks for the review. I fixed it!

tomoam:
I found that with this PR implementation, warnings are not displayed in case of compile errors.

https://user-images.githubusercontent.com/29677552/235648174-99cfcd58-fa88-4cc9-ad10-47d4682cda03.mov

This is because this PR implementation uses `onwarn` in svelte.config.js, which means that we are not notified that warnings have been cleared, so each time we clear the warnings store. (If using `sendWarningsToBrowser`, we are notified of an empty array.)

`onwarn` is line 140, `sendWarningsToBrowser` is line 156:
https://github.com/sveltejs/vite-plugin-svelte/blob/aba990bc8fd0dcaa8107ebf28e0504cdb0b376c9/packages/vite-plugin-svelte/src/utils/log.ts#L138-L156

If you want to fix this part too, please do not merge this PR yet, because I also need to change `vite-plugin-svelte`.

dummdidumm (approved):
It's fine that there's no warning when the file contains an error. VS Code and the current REPL behave the same. Thank you for fixing this!

PR #383, opened 2023-05-08
adding a style to button and remove unused css

JabSYsEmb:
`<button aria-label='Mark as complete'/>` is missing the style in the solution and delete unused .saving css style.


JabSYsEmb:
.saving class looks like it's added mistakenly to style tag in all 'Forms' exercises.
image

dummdidumm (approved):
Thank you!

Issue #306, opened 2023-04-02
WebContainers are in beta now for Safari


thedenoff:
Recent macOS update (13.3) bumped Safari to 16.4 which on [WebContainers Browser Support page](https://webcontainers.io/guides/browser-support#safari) isn't described as lacking some features anymore but now as "in beta".

WebContainer's landing page demo now successfully runs on Safari with `webcontainer_any_ua` in LocalStorage being set to any non-empty value
<img width="1374" alt="image" src="https://user-images.githubusercontent.com/13051154/229371249-ab208271-7c3f-46bc-9642-7c177d38c790.png">

I haven't though tested "learn"s code with removed user-agent matching codepath. Maybe there are some parts of WebContainers functionality which "learn" uses and which still don't work. If there are, maybe it would be smart to change check from user-agent to these specific function(s). At least we're getting close

Rich-Harris:
Feature-detection rather than UA sniffing would be ideal, yeah, though the docs aren't quite clear on how to do the feature detection (the 'such as' makes me nervous):

<img width="815" alt="image" src="https://user-images.githubusercontent.com/1162160/229374767-3bf63d98-e709-48e5-80b6-35ae3fbbfaa7.png">

I've opened a request for a) `boot()` to throw in unsupported environments and/or b) the ability to know ahead of time if the environment is supported: https://github.com/stackblitz/webcontainer-core/issues/1021

thedenoff:
Great, thank you for opening that request! Should I close then this one and watch there or leave this as is for now?

Rich-Harris:
Let's leave this open, as we'll still need to do some work on our end if/when they make any changes on theirs

thedenoff:
For those who really wants to learn.svelte.dev in Safari even _before_ this resolves: 

- it's possible in macOS 13.3 (or if you got Safari 16.4 in some other way), 
- make Safari pretend it's (for example) Chrome (in Dev menu) 
- tick "Disable Cross-Origin Restrictions" (in Dev menu too) 

<img width="1290" alt="image" src="https://user-images.githubusercontent.com/13051154/229379915-417bca13-4d40-418c-b03c-29516a2d4b5e.png">

thedenoff:
[...]

PR #385, opened 2023-05-10
chore: safari support

tomoam:
https://github.com/sveltejs/learn.svelte.dev/issues/306#issuecomment-1542155183

This PR is a temporary workaround, but useful for Safari users.
If using Safari 16.3 or earlier, you will see the same message as before.

### DIFF
--- a/src/routes/tutorial/[slug]/Loading.svelte
+++ b/src/routes/tutorial/[slug]/Loading.svelte
@@ -1,4 +1,6 @@
 <script>
+	import { isWebContainerSupported } from "$lib/client/adapters/webcontainer/utils.js";
+
 	/** @type {boolean} */
 	export let initial;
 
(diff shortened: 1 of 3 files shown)


src/routes/tutorial/[slug]/Loading.svelte
```diff
@@ -10,11 +10,16 @@
 
 	/** @type {string} */
 	export let status;
+
+	function webcontainer_is_supported() {
+		// @ts-ignore
+		return 'function' === typeof Atomics.waitAsync;
```
Comment (dummdidumm) [inline src/routes/tutorial/[slug]/Loading.svelte:16]: Can we be sure that `Atomics` exists, or do we need an extra check like `typeof Atomics !== 'undefined'`?
Also: Yoda speak 😄

Comment (tomoam) [reply src/routes/tutorial/[slug]/Loading.svelte:16]: You're right, this code would cause an error on browsers that don't support `Atomics`. But since https://github.com/sveltejs/learn.svelte.dev/issues/306#issuecomment-1538330568 gave me the production code from webcontainers.io, I would like to borrow it. I will therefore mark this as a solution.

dummdidumm (approved):

PR #388, opened 2023-05-10
Updating status code references

fubar1337:
Shorter domain, more status code support


dummdidumm:
What do you mean by "more status code support"? httpstatusdogs is well known and contains all relevant codes for this exercise.

Rich-Harris:
I've found myself using http.dog a bunch lately, so I'm in favour — thanks!

PR #398, opened 2023-06-02
Improve style of selected item in filetree

sir-kokabi:
I've changed the CSS of the filetree component to make the selected item stand out more from the background color. This helps users see and find the selected item better, especially if they have trouble with colors or vision.

### DIFF
--- a/src/routes/tutorial/[slug]/filetree/Item.svelte
+++ b/src/routes/tutorial/[slug]/filetree/Item.svelte
@@ -203,6 +203,7 @@
 		transform: translate(0, 0.2rem) rotate(45deg);
 		z-index: 2;
 	}
+	
 	[aria-current='true']:has(:focus-visible)::after,
 	:global(.mobile-filetree) [aria-current='true']::after {
 		display: none;
--- a/src/routes/tutorial/[slug]/filetree/Item.svelte
+++ b/src/routes/tutorial/[slug]/filetree/Item.svelte
@@ -188,7 +188,23 @@
 	}
 
 	[aria-current='true'] {
-		color: hsl(15, 100%, 50%);
+		color: var(--sk-theme-1);
 	}
 
+	[aria-current='true']::after {
+		content: '';
+		position: absolute;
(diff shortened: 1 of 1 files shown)


src/routes/tutorial/[slug]/filetree/Item.svelte
```diff
@@ -188,24 +188,7 @@
...
-		background-color: var(--sk-back-3);
-		border: 1px solid var(--sk-back-4);
-		transform: translate(0, 0.2rem) rotate(45deg);
-		z-index: 2;
-	}
-
-	[aria-current='true']:has(:focus-visible)::after,
-	:global(.mobile-filetree) [aria-current='true']::after {
-		display: none;
-	}
```
Comment (geoffrich) [inline src/routes/tutorial/[slug]/filetree/Item.svelte:210]: We shouldn't remove this - this adds the indented caret next to the selected file 
image

src/routes/tutorial/[slug]/filetree/Item.svelte
```diff
@@ -188,24 +188,7 @@
 	}
 
 	[aria-current='true'] {
-		color: var(--prime);
+		color: hsl(15, 100%, 50%);
```
Comment (geoffrich) [inline src/routes/tutorial/[slug]/filetree/Item.svelte:191]: Looks like the root issue here is that the `--prime` variable no longer exists - I think we should use `--sk-theme-1` instead of a hard-coded value

geoffrich (changes requested):

src/routes/tutorial/[slug]/filetree/Item.svelte
Comment (sir-kokabi) [reply src/routes/tutorial/[slug]/filetree/Item.svelte]: I have restored this section, but I believe the small arrow here is unnecessary. It is similar to making a menu item both bold and colorized.

src/routes/tutorial/[slug]/filetree/Item.svelte
Comment (sir-kokabi) [reply src/routes/tutorial/[slug]/filetree/Item.svelte]: I've change color to `--sk-theme-1`.

geoffrich (approved):

PR #402, opened 2023-06-08
Improved wording.

faetalize:
The wording provided initially is confusing, and I believe is mistaken.

### DIFF
--- a/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/README.md
+++ b/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/README.md
@@ -2,9 +2,9 @@
 title: Keyed each blocks
 ---
 
-By default, when you modify the value of an `each` block, it will add and remove items at the _end_ of the block, and update any values that have changed. That might not be what you want.
+By default, when you modify the value of an `each` block, such as adding or removing items, it will apply such modifications at the _end_ of the block, and then update any values that have changed. That might not be what you want.
 
-It's easier to show why than to explain. Click the 'Remove first thing' button a few times, and notice what happens: It removes the first `<Thing>` component, but the _last_ DOM node. Then it updates the `name` value in the remaining DOM nodes, but not the emoji, which in `Thing.svelte` is fixed when the component is created.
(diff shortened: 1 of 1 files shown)


content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/README.md
```diff
@@ -2,9 +2,9 @@
 title: Keyed each blocks
 ---
 
-By default, when you modify the value of an `each` block, it will add and remove items at the _end_ of the block, and update any values that have changed. That might not be what you want.
+By default, when you modify the value of an `each` block, such as adding or removing items, it will apply such modifications at the _end_ of the block, and then update any values that have changed. That might not be what you want.
 
-It's easier to show why than to explain. Click the 'Remove first thing' button a few times, and notice what happens: It removes the first `<Thing>` component, but the _last_ DOM node. Then it updates the `name` value in the remaining DOM nodes, but not the emoji, which in `Thing.svelte` is fixed when the component is created.
+It's easier to show why than to explain. Click the 'Remove first thing' button a few times, and notice what happens: It doesn't actually remove the first `<Thing>` component, but rather the _last_ DOM node. Then it updates the `name` value in the remaining DOM nodes, but not the emoji, which in `Thing.svelte` is fixed when the component is created.
```
geoffrich, line 7:
```suggestion
It's easier to show why than to explain. Click the 'Remove first thing' button a few times, and notice what happens: it does not remove the first `<Thing>` component, but rather the _last_ DOM node. Then it updates the `name` value in the remaining DOM nodes, but not the emoji, which is fixed when each `<Thing>` is created.
```

I more closely aligned this with a previous change on the old tutorial

content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/README.md
```diff
@@ -2,9 +2,9 @@
 title: Keyed each blocks
 ---
 
-By default, when you modify the value of an `each` block, it will add and remove items at the _end_ of the block, and update any values that have changed. That might not be what you want.
+By default, when you modify the value of an `each` block, such as adding or removing items, it will apply such modifications at the _end_ of the block, and then update any values that have changed. That might not be what you want.
```
geoffrich, line 5:
```suggestion
By default, when you modify the value of an `each` block, it will add and remove items at the _end_ of the block, and update any values that have changed. That might not be what you want.
```

I think this one reads better as is

geoffrich (commented):
Thanks! Looks like this was a change made to the old tutorial in https://github.com/sveltejs/svelte/pull/7218 that was missed

PR #403, opened 2023-06-08
docs: fix broken fragment in link

KyleMit:
Migrated from https://github.com/sveltejs/svelte/pull/8715

It looks like the following URL fragment doesn't exist anymore

<https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_bubbling_and_capture>

So while users used to get navigated to the section specifically on bubbling and capture - now they just see the entire Event page.  Which honestly, isn't bad reading, but the event page as a whole is relevant to pretty much every other item in the event modifiers section.  Instead, this inlines links directly to  sections on bubbling vs capture.


dummdidumm (approved):
Thank you

PR #414, opened 2023-06-28
Update README.md

thruthesky:
I think there is a mistake on the document.
Shouldn't it be `you must not return ...` instead of `you need to ...`


Conduitry:
No, what's there is correct. This is a list of situations where you won't be able to use a `+page.server.js` `load` function and will instead need to do something else.

PR #415, opened 2023-06-29
Add examples to @html tag warning

samal-rasmussen:
Whenever I have people who a new to web development try the Svelte tutorial, they always freeze at the ominous warning on the @html tag, because they have to wonder "what is HTML that comes from sources I don't trust?".

So I updated the text so it now uses a self authored article as an example of trusted content, and a comment on an article as an example of untrusted content. Hopefully this is a bit more approachable to new developers.


dummdidumm (approved):
Thank you!

PR #417, opened 2023-06-30
Update stores.js

gurelbs:


Conduitry:
This isn't what we want to demonstrate. `.update()` takes a function that inputs the old value and outputs the new value. Doing an in-place reassignment of a function parameter is misleading.

PR #419, opened 2023-06-30
Misplaced Bracket

flloschy:
Added a missing opening bracket


Conduitry:
Because we're using a dash here, I think we should instead get rid of the closing parenthesis, not add an opening one.

geoffrich (approved):

PR #425, opened 2023-07-13
Recognize that cookie value is a string

tamutus:
The tutorial presently makes it seem like a cookie's value can be a non-string. This is one possible adjustment to the exercise code to avoid such confusion. Alternatively, +page.server.js could get ```visited```'s value by performing the same check.


Rich-Harris:
thank you — I think we probably need to make `data.visited` a boolean in all cases (rather than going from `false` to `"true"`) so I'm going to close this in favour of #539

PR #429, opened 2023-07-15
Update README.md

Codooze:
Since it was not specific, so it oponed the question if the pogress variable would keep being a store.

It might be useful to know which other elements in Svelte Motion have similar behaviour to a store, as I understand some of these elements use store internally so making use of them in a variable makes that variable a store(?).


Rich-Harris:
thanks — i think we can solve this more easily, closing in favour of #540

PR #431, opened 2023-07-16
docs: change grammar to help reduce confusion

brokenthorn:
It was not clear which page below the layout inherits the data, because in fact it could be any page.

The confusion can arise from the way the previous sentence was expressed, as **visually**, there is nothing `below` the layout in the project tree.

There are siblings though, like `+page.svelte` but for a Svelte beginner it is **not** immediately obvious that this page also falls `below` (or **under**) the same layout.


Rich-Harris (approved):

Rich-Harris:
thanks!

PR #432, opened 2023-07-17
Update README.md

YegorKochetkov:
delete a duplicate example


geoffrich:
The example is not a duplicate - there are two forms on the page, one with `action="?/create"` and one with `action="?/delete"`

PR #437, opened 2023-07-18
docs: add missing verb for cookies.set() explanation

Zerotask:


Conduitry:
I think "you" -> "use" sounds better to me here than "you" -> "you call", but I'm not sure what sort of style we're using in the rest of this tutorial.

Zerotask:
> I think "you" -> "use" sounds better to me here than "you" -> "you call", but I'm not sure what sort of style we're using in the rest of this tutorial.

Personally I don't like the double "u".
In the following the verb "call" is used, e.g. https://github.com/sveltejs/learn.svelte.dev/blob/ee69014dfec7d05c434f3537ad263ca620a5f569/content/tutorial/03-sveltekit/04-headers-and-cookies/02-cookies/README.md?plain=1#L37

content/tutorial/03-sveltekit/04-headers-and-cookies/02-cookies/README.md
```diff
@@ -17,7 +17,7 @@ export function load(+++{ cookies }+++) {
 }
 ```
 
-To set a cookie, you `cookies.set(name, value, options)`. It's strongly recommended that you explicitly configure the `path` when setting a cookie, since browsers' default behaviour — somewhat uselessly — is to set the cookie on the parent of the current path.
+To set a cookie, you call `cookies.set(name, value, options)`. It's strongly recommended that you explicitly configure the `path` when setting a cookie, since browsers' default behaviour — somewhat uselessly — is to set the cookie on the parent of the current path.
```
dummdidumm, line 20:
```suggestion
To set a cookie, call `cookies.set(name, value, options)`. It's strongly recommended that you explicitly configure the `path` when setting a cookie, since browsers' default behaviour — somewhat uselessly — is to set the cookie on the parent of the current path.
```

dummdidumm (approved):

PR #442, opened 2023-07-19
docs: add info link for 422 status code

Zerotask:
I saw that this page https://learn.svelte.dev/tutorial/post-handlers pointed to https://httpstatusdogs.com/201-created. Therefore I decided to add a link to the [422 page](https://httpstatusdogs.com/422-unprocessable-entity).
Alternatively (probably more appropriate), the cat version: https://http.cat/status/422


dummdidumm (commented):
I feel like this info isn't really necessary, the focus of the tutorial is usage of `fail`, not what specific status code is used.

Rich-Harris:
I'm inclined to agree — this information is easily found, we shouldn't dilute the content. thank you

PR #443, opened 2023-07-19
Use fail fast for keydown handler instead of having an additional level

Zerotask:


dummdidumm (approved):
That's my prefered style, too, will wait on other maintainers to chime in before merging.

Rich-Harris:
thank you!

PR #447, opened 2023-07-20
Add definition order notice

Wombosvideo:
Add a notice telling the reader to make sure the definition order of stores and derived stores is correct.

## Describe the bug
Tauri won't load (blank window) the page if a `derived` store comes before the `writable` definition. For this reason, an addition to the tutorial page is suggested.

## Logs
```plaintext
Unhandled Promise Rejection: ReferenceError: Cannot access uninitialized variable.
```

### DIFF
--- a/content/tutorial/01-svelte/08-stores/04-derived-stores/README.md
+++ b/content/tutorial/01-svelte/08-stores/04-derived-stores/README.md
@@ -13,3 +13,5 @@ export const elapsed = derived(
 ```
 
 > It's possible to derive a store from multiple input stores, and to explicitly `set` a value instead of returning it (which is useful for deriving values asynchronously). Consult the [API reference](https://svelte.dev/docs#run-time-svelte-store-derived) for more information.
+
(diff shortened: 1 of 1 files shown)


Wombosvideo:
Referencing a variable `myArray` on a line number `x` less than the line number of the definition of `myArray` usually works without any problems, **as long as it was initialized before** running the code on line `x`.

```js
function sortArray() {
  myArray.sort((a, b) => {
    if (a < b) return -1;
    if (a > b) return 1;
    return 0;
  });
}
const myArray = [];
```

Since a derived store depends on the store it derives from (origin store), I assumed it would check if the origin store was initialized before trying to access its value.

My pull request was made because I spent over an hour trying to find out it was the definition order because it just wouldn't tell me **where in my code** the issue was, given my not so small codebase. You may be right signing this off as a rookie mistake from my side but I still believe others may benefit from the definition order notice.

Rich-Harris:
I'm inclined to agree with @Conduitry, this doesn't really belong here. This is just a fact about JavaScript

PR #448, opened 2023-07-22
docs: optimize verb for updating the button element

Zerotask:
Since the button element is already present in the code, one doesn't have to add a new button element but update the existing one.


Rich-Harris:
thanks — looks like this was updated elsewhere, closing

PR #452, opened 2023-07-28
Duplication of example code in tutorial of progressive-enhancement

minamme:
I think this is just a duplicate by mistake and it's confusing.


minamme:
It's my mistake I didn't notice action name difference.

PR #457, opened 2023-08-01
chore: align code solution with explanation

Zerotask:
code solution aligned with route groups explanation

https://github.com/sveltejs/learn.svelte.dev/blob/main/content/tutorial/04-advanced-sveltekit/04-advanced-routing/04-route-groups/README.md


dummdidumm:
Thank you, but #475 contained the more desired solution, so closing in favor of that.

PR #467, opened 2023-08-07
Replace `background` with `background-color` for consistency

baderj:
The tutorial on *Component styles* uses `background-color` in the README.md, while the app uses the shorthand `background`. This could be confusing:

1. The text tells us to 
   > change `background-color` ...
   
   but actually changes the `background` in `Box.svelte`.

2. The first snippet uses `background-color` to overwrite `background` in `App.svelte`.  

I replaced the shorthand `background` in the app with `background-color` to make it consistent with the tutorial text.


dummdidumm (approved):
Thank you!

PR #463, opened 2023-08-05
Fix sample code in tutorial text to match the exercise solution

gorisanson:
This commit fix sample code in the-form-element tutorial text, which is `01-the-form-element /README.md`, to match the exercise solution, which is `01-the-form-element/app-b/src/routes/+page.svelte`.

In [the-form-element tutorial page](https://learn.svelte.dev/tutorial/the-form-element),
I found that the sample code in the text was not matched the exercise solution as we can see in the following:

Sample code in the text:
```svelte
<h1>Todos</h1>

<form method="POST">
  <label>
    add a todo:
    <input
      name="description"
      autocomplete="off"
    />
  </label>
</form>

{#each data.todos as todo}
```

Exercise solution:
```svelte
<h1>todos</h1>

<form method="POST">
  <label>
    add a todo:
    <input
      name="description"
      autocomplete="off"
    />
  </label>
</form>

<ul class="todos">
  {#each data.todos as todo (todo.id)}
```

This PR fix this mismatch.

PR #469, opened 2023-08-08
Fixes inconsistincy in code and readme

philolo1:
The readme says "Todos", but the code says "todos", so i changed "todos" in the code.


dummdidumm:
Thank you, but #463 was the more correct fix, so closing in favor of that.

PR #471, opened 2023-08-08
Access store value in context example

jpaquim:
As an alternative, `console.log({ $count })` could also be used, not sure what's best for this example.


dummdidumm (approved):
Thank you!

PR #472, opened 2023-08-08
Update README.md

JonGracias:
Changed console.log('the count is ${count}');

to console.log('the count is', {count});

to correctly log the current count


Conduitry:
What's incorrect about how it's getting logged now?

The proposed change is definitely incorrect, because it has mismatched quote types.

dummdidumm:
These are backticks (`), not single quotes ('), so the current code is correct.

PR #475, opened 2023-08-09
Update route-groups doc to match target

jpaquim:
Component text and slot ordering should match the target app-b: https://github.com/sveltejs/learn.svelte.dev/blob/main/content/tutorial/04-advanced-sveltekit/04-advanced-routing/04-route-groups/app-b/src/routes/(authed)/%2Blayout.svelte


dummdidumm (approved):
Thank you!

PR #477, opened 2023-08-10
Fix missing initial state of `updated store` tutorial

salman-ar-sar:
This PRs fixes the initial state missing issue on Part 3: Basic SvelteKit > Stores > Updated store tutorial. 

It didn't have an initial state and stayed in the solved state.

### DIFF
--- a/content/tutorial/03-sveltekit/08-stores/03-updated-store/app-b/src/routes/+layout.svelte
+++ b/content/tutorial/03-sveltekit/08-stores/03-updated-store/app-b/src/routes/+layout.svelte
@@ -1,5 +1,5 @@
 <script>
-	import { page, navigating } from '$app/stores';
+	import { page, navigating, updated } from '$app/stores';
 </script>
 
 <nav>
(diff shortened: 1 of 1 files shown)


dummdidumm:
Tricky one - the problem with this tutorial is that it's not really possible to see the result as it would require a new version of `learn.svelte.dev` to be available. So this tutorial chapter just talks about how this looks like - but the git diff certainly makes it look like you would have to do something. Mhm..

salman-ar-sar:
> Tricky one - the problem with this tutorial is that it's not really possible to see the result as it would require a new version of `learn.svelte.dev` to be available. So this tutorial chapter just talks about how this looks like - but the git diff certainly makes it look like you would have to do something. Mhm..

Oh yeah that makes sense. But the tutorial shows the change and makes it look I should make that change. Comparing it with the previous tutorials made me think like that. 

But what you said makes sense, so what should we do?

Rich-Harris:
thank you. i don't think there's an ideal solution to this, but i think #544 is probably as close as we can get for now — it basically solves the 'what changed?' problem by inverting it. i'll close this in favour of that

Issue #424, opened 2023-07-12
On what license learn.svelte.dev is released ?


mojzesh:
Hey guys!

Thanks for creating Svelte and SvelteKit - It's awesome.

Could someone clarify on what LICENSE this repo is released ?
I couldn't find this information anywhere in the repo.

Thanks!

PR #479, opened 2023-08-11
chore: add license

mateusabelli:
This patch adds a MIT license based on SvelteKit [license](https://github.com/sveltejs/kit/blob/master/LICENSE). As suggested [here](https://github.com/sveltejs/learn.svelte.dev/issues/424#issuecomment-1633218028).
Pease let me know what do you think.

Fixes: #424


dummdidumm (approved):
Thank you!

PR #481, opened 2023-08-13
Remove `experimental` from inspector option

morinokami:
Update svelte.config.js to suppress the following warning:

> [vite-plugin-svelte] Option "vitePlugin.experimental.inspector" is no longer experimental and has moved to "vitePlugin.inspector". Please update your svelte config.


benmccann:
`holdMode` now defaults to `true`, so I updated it to remove this entire section, which is no longer needed

PR #488, opened 2023-08-21
docs: edit instruction for Store Bindings

glianx:
Edited instruction to 'add button' to 'add event handler', to reflect the true diff in the code. (The button already exists; the highlighted part is the event handler)


geoffrich (approved):
Thanks!

PR #495, opened 2023-09-01
docs: adds missing `readyState` binding

baderj:
Svelte added support for the `readyState` media element binding in https://github.com/sveltejs/svelte/pull/6843 . While the docs have been updated to include this option, it was missing in the tutorial.


dummdidumm (approved):
Thank you!

PR #500, opened 2023-09-11
docs: update solution

c31io:
Sync the solution with the step-by-step guide.


Rich-Harris:
the step-by-step guide _does_ end with `{name} dances`:

image

you have to know that the man is called 'Rick Astley' but that's part of the joke :)

PR #501, opened 2023-09-12
Disabling button also when selected is undefined

pernyblom:
I don't know if this is a good practice in this example, but it makes sense if the options are fetched from a server


pernyblom:
Just realized that there are two files that need to be changed (the solution and the original)

Rich-Harris:
thanks. i think it'd be a bit confusing in this case, because `selected` is always populated once the component mounts, so i'll close this

PR #504, opened 2023-09-16
fix: restart audio playback on repeated button pushes

elias-pap:
When the button is clicked for a second time before the audio playback triggered by the first click is finished, the second click is ignored. Restarting the playback on every click makes the button feel more responsive.
Useful e.g. when the button is clicked very fast, many times in a row.


Rich-Harris (approved):

Rich-Harris:
thanks!

PR #508, opened 2023-09-25
Assign the ID back when creating a new userid in cookies

HendrikPetertje:
Thanks for making this amazing tutorial, it's so nice!

While following the tutorial I noticed that one excersise would cause a 500 to be rendered the first time the code was loaded, as I made just one change the page would re-render and all would be fine.

Upon looking closer I noticed that map get was being called with `undefined`. 

This little change makes sure db.getTodos(id) isn't called with undefined.


HendrikPetertje:
oops.

Rich-Harris:
thanks! there's a few more places that need updating, so i'll close this in favour of #543

PR #516, opened 2023-10-11
Update README.md

LWChris:
Adjust source code snippet according to modifications from previous step.

In the previous step, the tutorial asks the reader to change `alt="A man dances."` into `alt="{name} dances."`. The code snippet should reflect that change.


dummdidumm (approved):
Thank you!

PR #517, opened 2023-10-12
Sidebar: scroll to top when exercise change

sorae42:
The current `scrollTop` doesn't work, but this one work with Chrome and Safari. I haven't tested on non-Chrome browsers yet.


Rich-Harris:
thank you — this was also fixed in #530, so I'll close this

PR #524, opened 2023-11-02
Update README.md to change the word truthy to true

MarwanElKhodary:
Changed truthy to true


Conduitry:
"Truthy" is the intended word here. https://developer.mozilla.org/en-US/docs/Glossary/Truthy

PR #525, opened 2023-11-09
Fix: sidebar scrolls to top after navigate

snowiewdev:
Fix for sidebar should scroll to top after navigate:

Current Issue:
The current sidebar in tutorial doesn't scroll back to top when clicking `next` tutorial. 
User needs to scroll back to top manually, which is not convenient, e.g.
- Clicking `Next: tick` on sidebar in [https://learn.svelte.dev/tutorial/update](https://learn.svelte.dev/tutorial/update), the scrollbar in sidebar doesn't get scrolled to top in next tutorial

This fix should resolve the issue.


Rich-Harris:
thank you — this was also fixed in #530, so I'll close this

PR #534, opened 2023-12-10
fix: replace `:` with `,`

jrwiegand:


Conduitry:
`:` is fine.

jrwiegand:
@Conduitry
If you say so.
¯\_(ツ)_/¯

PR #537, opened 2023-12-12
docs: Clarify usage of `tick` function in Part1/Lifecycle/Tick section

abyrne10:
## Purpose of This Pull Request

This PR updates the documentation for the `tick` function in the lifecycle section of the Svelte tutorial. The goal is to clarify the unique behaviour of `tick` compared to other lifecycle functions like `beforeUpdate` and `afterUpdate`.

## Key Changes Made

Revised the description to emphasise that `tick` can be manually called at any point in the component's code. This distinction is important as it differentiates tick from other lifecycle hooks that are automatically invoked by the Svelte framework.

Provided additional context to explain that `tick` is used for synchronising with the DOM update cycle after state changes, unlike the automatic nature of lifecycle hooks which are triggered by the framework at specific times.

## Rationale for Changes

The previous documentation mentioned that the `tick` function can be called "not just when the component first initialises." This phrasing could potentially be misunderstood to imply that other lifecycle functions are only called during initialisation, which is not the case. The revised wording aims to eliminate this ambiguity, presenting a clearer understanding of the `tick` function's role in Svelte's reactivity system.

## Impact of Changes

The changes made in this PR should provide readers with a more accurate and comprehensive understanding of how and when to use the `tick` function effectively within Svelte components. This clarity will aid developers in writing more efficient and understandable Svelte code.

### DIFF
--- a/content/tutorial/01-svelte/07-lifecycle/03-tick/README.md
+++ b/content/tutorial/01-svelte/07-lifecycle/03-tick/README.md
@@ -2,7 +2,7 @@
 title: tick
 ---
 
-The `tick` function is unlike other lifecycle functions in that you can call it any time, not just when the component first initialises. It returns a promise that resolves as soon as any pending state changes have been applied to the DOM (or immediately, if there are no pending state changes).
+The `tick` function is unique among Svelte's lifecycle tools in that it can be manually called at any point in your component's code, rather than being automatically invoked by the framework. It returns a promise that resolves as soon as any pending state changes have been applied to the DOM, or immediately if there are no pending state changes.
 
 When you update component state in Svelte, it doesn't update the DOM immediately. Instead, it waits until the next _microtask_ to see if there are any other changes that need to be applied, including in other components. Doing so avoids unnecessary work and allows the browser to batch things more effectively.
 


dummdidumm:
The current text is better because it's more accurate: you call these lifecycle methods, and they then register themselves somewhere which then invokes something at the appropriate time - so yes, the frameworks does call something at a specific point in time, but the actual visible call is still done by you, the developer. Therefore closing - thank you.

PS: This AI generated PR description is pretty lengthy and duplicative, makes it harder to understand what was actually changed.

abyrne10:
> The current text is better because it's more accurate: you call these lifecycle methods, and they then register themselves somewhere which then invokes something at the appropriate time - so yes, the frameworks does call something at a specific point in time, but the actual visible call is still done by you, the developer. Therefore closing - thank you.
> 
> PS: This AI generated PR description is pretty lengthy and duplicative, makes it harder to understand what was actually changed.

Thank you for your reply and your feedback. I take your point that the developer calls the lifecycle methods to register a callback that then gets called by the framework, rather than the framework calling the lifecycle methods themselves.

Can I just check my understanding is correct? The difference between `tick` and, for example, `beforeUpdate` is that if you call `beforeUpdate` to register a callback outside of the component initialisation (e.g. inside a `handleKeydown` function), the registration will not succeed?

PR #545, opened 2023-12-17
fix(sveltekit/6-forms/lesson-1): bug in solution

h-sifat:
The solution of the lesson **"The \<form\> element"** crashes because of an assignment to a constant.


taylorgorman:
Recommend #546 instead of this as it resolves other instances of the same issue.

geoffrich:
Thanks, merged #546 instead since it's a more comprehensive change.

PR #546, opened 2023-12-17
fix: const to let in "the <form> element"

golfomania:
**Lesson**
[The Form Element](https://learn.svelte.dev/tutorial/the-form-element)

**Problem**
In the load function in +page.server.js a new userid is created if not found in cookies
But id is declared as const so it can not be updated
Tutorial app run into 500 internal error

**How to reproduce**
e.g. open link in incognito window

**Fix**
change declaration from id from const to let

Martin
_This is my first open source contribution, so i hope this PR works for you this way_

### DIFF
--- a/content/tutorial/03-sveltekit/06-forms/01-the-form-element/app-a/src/routes/+page.server.js
+++ b/content/tutorial/03-sveltekit/06-forms/01-the-form-element/app-a/src/routes/+page.server.js
@@ -1,7 +1,7 @@
 import * as db from '$lib/server/database.js';
 
 export function load({ cookies }) {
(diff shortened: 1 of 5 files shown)


golfomania:
found more sections with the same code and changed it as well in second commit

golfomania:
related to #543

JonathanHarford:
I was just now working through the tutorial and bumped up against that problem. Thanks!

golfomania:
@JonathanHarford as long as the PR is not accepted just fix it manually in the file 👍

taylorgorman:
Came here to fix this, happy to see you beat me to it!

PR #548, opened 2023-12-27
Fix duplicated code snippet

truongnmt:
### Why
Correct me if I'm wrong but currently in the `Part 3 / Forms / Progressive enhancement` part, the instruction code snippet seems duplicated.
https://learn.svelte.dev/tutorial/progressive-enhancement

image

### What
Remove the duplicated code snippet.


geoffrich:
This is expected - the `action` is different between the two snippets. You're supposed to enhance the create and delete forms.

PR #550, opened 2023-12-28
docs: adjust div class value for consistency

kshyun28:
While going through the [Slots](https://learn.svelte.dev/tutorial/slots) section, I noticed the values of the `div` class were inconsistent.

The text has a space in the value, `<div class="card ">`.

While the `Card.svelte` file didn't have any spaces, `<div class="card">`.

### Tutorial Text
Card.svelte code snippet

### Tutorial File `Card.svelte`
Card.svelte source file


geoffrich (approved):
Thanks!

PR #551, opened 2023-12-28
docs: add code snippet for removing if blocks

kshyun28:
In the [svelte:component](https://learn.svelte.dev/tutorial/svelte-component) and [svelte:element](https://learn.svelte.dev/tutorial/svelte-element) sections, the tutorial is missing the step to remove the `if blocks` after adding `<svelte:component>` and `<svelte:element>` respectively.

This PR adds a step to `remove if blocks` in the shown code snippets.

## Behavior
- Following the current state of the tutorial doesn't `solve` the problem.
- The problem is only `solved` after removing the `if blocks`, which the tutorial does not indicate.
- There is a duplicate element as shown in the images below, since we don't remove the `if blocks`.

## Images
Observe the following from the images:
- code snippet for `App.svelte` file.
- output of the source files, which displays the additional element from the `if block`.

### <svelte:component>
<svelte:component>

### <svelte:element>
<svelte:element>

### DIFF
--- a/content/tutorial/02-advanced-svelte/09-special-elements/02-svelte-component/README.md
+++ b/content/tutorial/02-advanced-svelte/09-special-elements/02-svelte-component/README.md
@@ -28,6 +28,12 @@ We _could_ do this with a sequence of `if` blocks...
 </select>
 
 +++<svelte:component this={selected.component}/>+++
+
+---{#if selected.color === 'red'}
+	<RedThing />
+{:else}
+	<p>TODO others</p>
+{/if}---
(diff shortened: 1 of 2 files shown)


kshyun28:
I see, thank you for clarifying @Rich-Harris! Will keep this in mind moving forward. 

I also feel having a document outlining guidelines for making tutorial content would be helpful (if there isn't any already).

P.S. Really enjoyed the interactive tutorials! Now just looking to apply what I've learned for my own use cases.

PR #554, opened 2024-01-01
Solve the ARIA-ROLE error thrown @Part 2/Actions/The use directive

Ayyanaruto:
Screenshot from 2024-01-01 12-55-02

Added a role attribute to remove the warning in both unsolved and solved one


Rich-Harris:
thank you! @geoffrich can I ask you to take a look at this — is `presentation` the right role, or would `button` be more correct in this case?

Rich-Harris:
Oh and there's also #567 which additionally adds an `aria-label` — is that appropriate in this context?

geoffrich:
Hm, I think this is actually a case where it's perfectly valid to ignore the warning. Maybe we want to add the `svelte-ignore` comment to this example?

In general, you don't want to put click handlers on divs. However, in this case a screen reader can close the modal with the Escape key, so we aren't locking any functionality behind clicking with a mouse.

I'd have to test with a screen reader to be sure, but I don't think `role="presentation"` will actually do anything tangible other than silence Svelte's warning. A `div` is already meaningless, semantically.

PR #566, opened 2024-01-20
Inform user to install marked for local usage

furnivall:
Apologies if this has been mentioned in the past, but I figured it was worth letting people know that they'll need to install the `marked` dependency if following the tutorial locally.

This is based on the presumption that they may not understand it's not a Svelte built-in feature.

This is not mentioned in the existing documentation.

### DIFF
--- a/content/tutorial/01-svelte/06-bindings/07-textarea-inputs/README.md
+++ b/content/tutorial/01-svelte/06-bindings/07-textarea-inputs/README.md
@@ -2,6 +2,8 @@
 title: Textarea inputs
 ---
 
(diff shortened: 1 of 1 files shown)


dummdidumm:
The tutorials are meant to be used online, such a sentence would confuse the vast majority of people. I think such a statement makes more sense inside the top level readme of this project

Rich-Harris:
Yeah, my inclination is to close this. If someone is following along locally it means they're sufficiently proficient to have set up a project, installed dependencies, structured it in such a way that they have an `App.svelte` to work inside, and started a dev server via the command line. It's vanishingly unlikely that someone would have the knowledge to do all that but not understand that you need to install dependencies before you can use them. Thanks anyway

PR #567, opened 2024-01-21
Update App.svelte

anandbmuley:
Added aria role and label for the main modal as it was failing with some error. Apart from the error it will make the app more accessible.


anandbmuley:
@Rich-Harris  Thanks for resolving it.

PR #573, opened 2024-02-03
fix(03-sveltekit/04-headers-and-cookies/01): add x-content-type-options header

### LINKED ISSUE #476: Sveltekit tutorial spams downloads on Firefox
On: https://learn.svelte.dev/tutorial/headers

When you're typing out the code in the instructions, once you get to 'text/plain' the browser will start spamming downloads of the contents in the iFrame (i.e. if you have "this is a bug" it will download a file with that text in it). 

I couldn't replicate this in Chrome

Replication:

copy and paste this into src/routes/+page.server.js

export function load({setHeaders}) {
	// set headers
	setHeaders({
		'Content-Type': 'text'
	})
}

In my browser, the page begins reloading constantly at this point.
If you add a '/' at the end of 'text' it will start downloading constantly. 

I've noticed that if I paste the entire solution, or preempt the setHeaders({...}) part with a return after it, the issue doesn't happen.

seqshem:
fixes #476

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options

### DIFF
--- a/content/tutorial/03-sveltekit/04-headers-and-cookies/01-headers/README.md
+++ b/content/tutorial/03-sveltekit/04-headers-and-cookies/01-headers/README.md
@@ -10,6 +10,7 @@ Most commonly, you'd use it to customise caching behaviour with the [`Cache-Cont
 /// file: src/routes/+page.server.js
 export function load(+++{ setHeaders }+++) {
 +++	setHeaders({
+		'X-Content-Type-Options': 'nosniff',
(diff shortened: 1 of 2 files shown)


seqshem:
I don't love adding a header in the tutorial that only applies to some browsers, so maybe this example could be re-worked, but at least this prevents the behavior mentioned in #476

PR #502, opened 2023-09-13
Update paper.svg fill attr

joelinman-nxp:
in the tutorial excercise 'named slots', the 'paper.svg' has fill="none". The exercise is to build something that looks like a 'business card' on a table it seems, however you cannot see the card, so it looks more like words etched into wood! screenshot : https://prnt.sc/kpg6vljbZT-T

PR #582, opened 2024-02-15
fix(02-advanced-svelte/07-composition): fix invisible `paper.svg` background

noelforte:
Caught this while going through the tutorial. When used as a background image, not including a fill will make the svg invisible, this commit adds the fill so the example works properly. Have also adjusted the noise `surfaceScale` so the effect is a bit more subtle.

Tested in local dev environment to ensure this works as intended.


noelforte:
Had to check to make sure I wasn't duplicating someone else's work after I made these changes, or if there was some previously raised issue that this pull would close, and came across #502. Seems that there were a few places this issue was still causing problems. Since I adjusted the appearance of the svg in `advanced-svelte/07-01` I thought I'd carry my changes over to the rest of the assets in the other tutorials. Feel free to undo that if it's not an appropriate change to make. Cheers!

PR #583, opened 2024-02-17
fix(svelt:events): ensure m prop is synced

pedroyremolo:
# Description

After solving the tutorial exercise of section Part 1. BasicSvelt > Events > DOM Events as requested by the tutorial, the pointer position was not being updated at the DOM. This was happening because the m object was not being assigned after changes, this PR solved it.

> Disclaim: Opted for `m = m;` instead of whole object assignment because the next section already suggests it


pedroyremolo:
You're right mate, don't remember the scenario anymore but the rendering was not happening unless I forcefully the `m = m` change.

But since I cannot simulate the issue anymore, all good in the neighborhood. Thanks for the support

PR #594, opened 2024-03-02
fix: add missing semicolon

emmyrlim:
altho it's generally unimportant, there is missing a semicolon here.


eltigerchino:
Thanks!

PR #596, opened 2024-03-17
typo

geo4orce:


Conduitry:
"open" is acceptable as an adjective here, and I think it sounds better than "opened".

PR #598, opened 2024-03-23
Update README.md

javiersuzuki:
Removed additional `$` sign from `$$props` in Note if need to reference all the props that were passed into a component.


Conduitry:
`$$props` is correct. This tutorial is for Svelte 4, and has not been updated for 5, if that's what the confusion is.

PR #609, opened 2024-05-05
Fix crash in tuto

pikooli:
Hello, 
Here a small fix in the tuto on hook. 
It was passing `stack` that was undefined to the console.error. 
So it crash.


Siv-tspab:
Up
It is still broken, and this fix would still fix the issue :eyes:

PR #611, opened 2024-05-12
fix: small typo 'ssr' -> 'csr'

Skyrdow:
this is a one character change in the tutorial


Conduitry:
This isn't a typo. It's saying, similarly to how you can use the `ssr` page option to disable server-side rendering (mentioned in the previous section), you can also use `csr` to disable client-side rendering.

PR #614, opened 2024-05-14
Fix ZIP Code typo in slots examples

uckelman:
The slot examples have the ZIP Code on the business card as `100099`, which is too many digits for a ZIP Code and not the one on the card in the film. This PR corrects them to `10099`.


Rich-Harris (approved):

Rich-Harris:
whoops, thanks!

PR #626, opened 2024-05-30
Fix spin function syntax for Svelte tutorial

musrex:
Corrected arrow function syntax to include missing parenthesis around the "t" parameter and added semicolon after the last backtick in returned CSS string. The auto-grader expects both.


Rich-Harris:
thank you!

PR #628, opened 2024-06-04
Update else-blocks example so that it makes sense when {count} is 0 or 10.

srenderfrance:
Currently the <p> that is rendered by the else statement is false when count is 0 and 10. It says "0 is between 0 and 10" and when count reaches 10 it says "10 is between 0 and 10". The proposed change results in true statements for all numbers that are not greater than 10.


Conduitry:
What's there now sounds more natural to me, and, given that the count starts at 0 and can only ever increases, it's also still correct.

PR #634, opened 2024-06-23
Misplaced comment from part 4 06-01

Jfhseh:
The comment "// if FEATURE_FLAG_X is not enabled" should be placed outside the if statement, since the if statement is checking if the feature flag is enabled.
image


Jfhseh:
image

Conduitry:
What's there now is correct. It's a single sentence - `code in here will be removed from the build output if FEATURE_FLAG_X is not enabled`.

PR #647, opened 2024-08-27
Update README.md

Luca-Mezger:
fixed typo: added "y" for y coordinate


Conduitry:
The 'x' here is a times/cross, not the letter 'x' to be accompanied by the letter 'y'.

